PR #12, opened 2022-01-24
Code refactor (with bug fixes)

sanguinariojoe:
* Fixed wrong index in GetNodePos()
* Return more intelligible errors in MoorDynInit(), avoiding to throw exceptions (since it is an C API)
* Added doxygen documentation to MoorDynInit()
* Improved the Makefile (Just for Linux (so lib)). Anyway CMake building tools can be considered in the future
* Added some basic infrastructure for testing, which has been of use to run Valgrind
* Improved the gitignore
* Fixed several memory errors, many related with uninitialized variables
* Fixed errors when MoorDyn is used several times in the same process
* Fixed some invalid strings treatments
* No more -w to suppress warnings during the compilation
* The Moordyn system is now packaged in a class
* A new API which makes possible to deal with different systems in the same process
* Introducing namespaces
* Some variables are conveniently named, like error codes and connection types
* Now it is possible to cross-compile the DLL from Linux (using MinGW 32/64)
* Now it is possible to cross-compile the DYLIB from Linux (using clang++)


sanguinariojoe:
I think these bunch of changes are ready to be merged @mattEhall 

Please, let me know!

sanguinariojoe:
Hey @mattEhall !

After this last bunch of changes I really think the cod looks quite better. On top of that I fixed several bugs. I would really appreciate if you can merge this, so we work in a common code base

mattEhall:
Hi @sanguinariojoe, this looks like amazing work. I'm really impressed. But I wish you'd gotten in touch with me during the process. I'm a bit daunted to see ~6000 lines of code changes out of the blue. Although I'm sure it has significant improvements and looks much higher quality than my coding, I'd need to get familiar with what's changed and evaluate any implications for ongoing work at NREL before potentially merging in all these changes. And, as you can guess from my delay, I don't have very much time to devote to this unfortunately. Could you contact me at my NREL email to talk more?
Thanks for what looks like a very valuable and substantial contribution!

PR #17, opened 2022-03-01
Added Pre-Compiled Files

TGiles1998:
Hi,

I have added the pre-compiled files from the WEC-SIM folder for easy access and use so that some users won't have to compile the software themselves. Hope this is okay!

many thanks,

Thomas


sanguinariojoe:
My 2 cents: This is not a good idea but better adding [Releases](https://github.com/mattEhall/MoorDyn/releases)

mattEhall:
Thomas, thanks for this. 
I think this is something removed in a prior pull request - @sanguinariojoe, what's your reason again?

I'd lean toward prioritizing convenience/access (by including binaries) over worry about best practices at this point...

sanguinariojoe:
Well, first I should mention I removed the binaries from the dev branch, this PR points towards master.

Regarding binaries in git, in general you DO NOT want to track them. Along this line we can differentiate 2 kinds of binaries:

 - Images or similar resources. Those binaries would not change, ever
 - Compiled binaries and automatically generated files, that will change at every single commit (or they will become obsolete, which is even worst)

The former can be safely added to the repo while the latter is the source of many problems. First, git has to track every single bite changed at every single commit. With time this will make the git repo giant and sluggish. Second, the more "random" changes you introduce the higher chances you have to create a hash collision, rendering the git repository broken forever (this is anyway very unlikely to happen). Finally, it is making branches and forks useless, e.g.

 1. I fork the project
 2. I make changes, so I recompile and add the library
 3. Meanwhile @mattEhall makes some other changes, obviously adding a different version of the compiled library
 4. I submit my PR and it cannot be merged because of conflicts.

Might be the case (or not) that there are conflicts to be fixed in the source code files, and they would be indeed hard to fix, but manageable anyway. But how can anyone fix the conflicts on the compiled library??? There is no way to workaround such problem, it is mandatory to start the work all over again. So bad...

mattEhall:
@TGiles1998, I think @sanguinariojoe has made a really solid case here, and I should add those binaries into a release rather than merging this pull request. Let's leave this open until I get it done!
Thanks to both of you,
Matt

sanguinariojoe:
Can we close this?

Just for the record... The v2 candidate in https://github.com/mattEhall/MoorDyn/pull/31 will autodeploy releases with Windows installers. And example here:

https://github.com/sanguinariojoe/MoorDyn/releases/tag/nightly

(And installing in Python will be as simple as typing `pip install moordyn`)

relhamoud:
Hi @TGiles1998 

I am struggling to get MATLAB to talk to MoorDyn. Here is an image that shows my script that i copied from MoorDyn user's guide.


Capture

And as i have read, the mooring subdirectory contains a text file named lines.txt

However, i am getting the following error.


Capture_2

Could you please help me ?

Best Regards

sanguinariojoe:
Please @relhamoud , DO NOT NECROBUMP issues.

On top of that, this issue has been discussed many times, and the reply is always use the same: Use MoorDyn v2, it can be compiled with a Matlab wrapper

sanguinariojoe:
Feel free to open a new issue if you have further problems

PR #19, opened 2022-03-03
WIP: New API

sanguinariojoe:
This is the last step to get what I think will be the ultimate API :-) . For the time being I am just addressing the Connection object, but the plan is extending it to the other objects (if we come to an agreement), producing a HUGE Pull Request.

As you can see, I renamed Connection.h as Connection.hpp, since it contains the C++ API (the actual C++ object). Then I created a new Connection.h containing the pure C API.

To avoid names collisions, all the C API methods are named MoorDyn_\*, and all the abstract types MoorDyn\*. Along the same line, the C++ objects are moved into the namespace moordyn::.

Thus 3rd parties and users might now:

 - Use the legacy C API, including MoorDyn.h (the functions are renamed, maybe we want to add a bunch of definitions so the users do not need to edit a single line of code?). This option is just redirecting the work to the new C API, helped by a singleton
 - Use the new C API, including MoorDyn2.h. That file is including all the subobjects C APIs (for the time being only Connection.h), without exposing a single line of C++. It is more clean, since all the methods affecting the connections are packaged in Connection.h, and they can enjoy the abstract type MoorDynConnection.
 - Use the C++ API, including MoorDyn2.hpp. That is giving full control of the library.

I would do (upon agreement) the same job for all the header files of the objects (Body, Line, Rod, Waves), i.e. renaming the .h as .hpp, moving the c++ object inside the moordyn:: namespace, and adding a .h with the C API which has an abstract type to handle each object. I am of course meanwhile including all the objects in the error and logging handling system.

Bonus track: I added the doxygen documentation for Connection (and it is becoming beautiful!). The other objects will be documented as well. This way you do not need to document the API in the .rst files, you can just link _build/html/group__old__c__api.html and _build/html/group__new__c__api.html


mattEhall:
Interesting. I'm still working on wrapping my head around everything. Thanks for the explanation and discussion. It's very helpful. I will try to understand more this week.

In your above message, I couldn't understand what you meant by the following:
- "and all the abstract types MoorDyn"
- "maybe we want to add a bunch of definitions so the users do not need to edit a single line of code"

Could you elaborate on these?

Also, I noticed that previously you changed the names of some functions when putting them in the new API:
- MoorDyn_**InitExtWaves** vs **ExternalWaveKinInit**
- MoorDyn_GetWavesCoords vs GetWaveKinCoord**inate**s
- MoorDyn_SetWave**s** vs SetWave**Kin**

This subjective renaming adds complication without any benefit, so let's aim for consistent names going forward. If things deserve better names, we can plan out a consistent renaming across the board. By the way, I'm planning on renaming all Connections to Points at some point in the future, once the dust is settled on all the restructuring.

I'd like to better understand what you have in mind for the logging. I had begun a log file capability, which you may have noticed in a few places (controlled by env.writeLog). Is the logging you have in mind similar to this (just a more sophisticated approach)? Or would it do something different?

I'm looking forward to checking out the start of the Doxygen documentation! This was always too far away on my to-do list. It will be wonderful to have.

sanguinariojoe:
> and all the abstract types MoorDyn

Ops! Markdown trolled me, I meant MoorDyn\* (it is fixed now in the comment above). There are 2 ideas convoluted here (my bad):

 - In the API you absolutely want to have a unique prefix. In this case I am using MoorDyn_\* for the C functions, Moordyn\* for the C abstract types, and moordyn:: namespace for C++ stuff. That way you will never have a names collision when someone links with many other libraries. I feel like you already had this problem before, because in the master branch you are exposing C methods named LinesInit(), a name that can be easily considered by some other libraries. Thus, linking MoorDyn and that other library in the same binary will be impossible
 - In the C API you absolutely want to identify every object with an abstract type, i.e. a thingy the user might use as identifier at the C API functions, but cannot inspect inside. Right now they are defined as `typedef void* MoorDyn` and `typedef void* MoorDynConnection`, i.e. they both boil down to a `void*` type. But the idea is to have some intermediate struct, so the user might not use the wrong instance in the wrong method (not at least accidentally). Something like `typedef struct _MoorDyn* MoorDyn` (where `struct _MoorDyn` is defined in MoorDyn.cpp, so the C type remains abstract)

> "maybe we want to add a bunch of definitions so the users do not need to edit a single line of code"

Yeah, the point is that in MoorDyn v1 (master branch) the methods were called `LinesInit()`, `LinesCalc()`... While now they are called `MoorDynInit()`, `MoorDynStep()`. So maybe we want to add some definitions like

```
#define LinesInit MoorDynInit
#define LinesCalc MoorDynStep
```

They can be protected by some `#ifndef NO_LEGACY_API`, so the user might always disable them, to avoid the naming conflicts discussed above, or they can keep them doing nothing, i.e. he can link with the new version of the library without needing to change his code (he would need to edit the input files though). I do not know if it is worthy...

> Also, I noticed that previously you changed the names of some functions when putting them in the new API:
> 
>     MoorDyn_InitExtWaves vs ExternalWaveKinInit
>     MoorDyn_GetWavesCoords vs GetWaveKinCoordinates
>     MoorDyn_SetWaves vs SetWaveKin

My bad! I suppose the most convenient approach is to just add the MoorDyn_ prefix, i.e.

 - MoorDyn_ExternalWaveKinInit
 - MoorDyn_GetWaveKinCoordinates
 - MoorDyn_SetWaveKin

Let me know

> I'd like to better understand what you have in mind for the logging. I had begun a log file capability, which you may have noticed in a few places (controlled by env.writeLog). Is the logging you have in mind similar to this (just a more sophisticated approach)? Or would it do something different?

The logging to a file and the logging to screen shall be clearly unified. The most straight forward approach is to just do not print a file, and let the user redirect the output (you can do that while launching an executable). We can also make the logging print to screen as well as to a file simultaneously.

For the time being I am just working on the screen side, and we can decide how to make it in a file too.

It will be controlled by the `MoorDyn_SetVerbosity()` function, and optionally in the input file (I do not like it though).

> I'm looking forward to checking out the start of the Doxygen documentation! This was always too far away on my to-do list. It will be wonderful to have.

It is beautiful! And it will be even more beatiful

source/Body.cpp
```diff
@@ -91,7 +90,7 @@ void Body::setup(int number_in, types type_in, double r6_in[6], double rCG_in[3]
 
 
 // this function handles assigning a line to a connection node
-void Body::addConnectionToBody(Connection *theConnection, double coords[3])
+void Body::addConnectionToBody(moordyn::Connection *theConnection, double coords[3])
```
sanguinariojoe, line 93: > Namepaces are great! We should use them more

Stupid quote to demonstrate commits commenting capabilities

sanguinariojoe:
> Yep, for the API names, just having MoorDyn_ be a prefix to existing names would be great.

Done [here](https://github.com/mattEhall/MoorDyn/pull/19/commits/6f2a30536496b4354f2fbe2d1b8a7d16308a4f5a)

> I'll just say that I get nervous about adding complexity, so the typedef struct _MoorDyn* MoorDyn approach sounds a little scary.

The C code is always packed with those little monsters... What about the following examples?

```
#ifndef __MOORDYN2_H__
#define __MOORDYN2_H__
...
#endif
```

```
#ifdef __cplusplus
extern "C"
{
#endif
...
#ifdef __cplusplus
}
#endif
```

They are needed... It is a standard everyone accepts (In modern C the former is actually replaced by `#pragma once`, but I am keeping the former standard all along the C API). Using those `typedef struct _MoorDyn* MoorDyn` it is also a standard, e.g. https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl.h#L29 .

The underlying idea is that you cannot accidentally call the wrong method with the wrong object, i.e.

```
auto system = MoorDyn_Create("../../test/Mooring/lines.txt");
auto conn = MoorDyn_GetConnection(system, 1);
MoorDyn_Init(system, NULL, NULL);  // OK
MoorDyn_Init(conn, NULL, NULL);  // Compilation error
```

Maximum control, zero risk, totally transparent to the user. Let me know if I convinced you so far

> About the abstract types, do we really need to give the user a handle for every object? Wouldn't it be easier/simpler to have the user just interact with the top-level MoorDyn object, using functions that direct to the desired Line/Connection/etc?
> E.g. getConnectionPosition(int connID, double rOut[3])
> Let's talk more about the use cases for needing access to objects in MoorDyn.

I do not think so. In your approach the user is using 2 identifiers at each call to the connection related methods:

```
auto system = MoorDyn_Create("../../test/Mooring/lines.txt");
int conn = 1;  // This is your conn identifier
double pos[3];
MoorDyn_GetConnectionPosition(system, conn, pos);
```

In my approach he only uses one:

```
auto system = MoorDyn_Create("../../test/Mooring/lines.txt");
auto conn = MoorDyn_GetConnection(system, 1);  // The identifier is the abstract type
double pos[3];
MoorDyn_GetConnectionPosition(conn, pos);
```

So in terms of user usage there are not very much benefits, are they? There are however some drawbacks:

 - If the identifier for every all objects (Lines, Nodes, Connections, Rods, ...) is always an integer, they can be messed up
 - Every single time you ask for info about an object you must check that the identifier is valid, reducing the performance (in an unnoticeable way, let's face that)
 - Many other libraries use the latter approach, so I really think the majority of users would feel more comfortable with that

For the sake of devs (you and me), you should really prefer the new approach. Now everything related to the connections can be moved from MoorDyn2.h and MoorDyn2.cpp to Connection.h and Connection.cpp. That way getting more methods in the C API would not result in huge MoorDyn2.h and MoorDyn2.cpp files.

Should I proceed?

> I'm not too worried about users having to adjust code from v1 to v2. There are such significant changes between the versions, it's maybe helpful to not let user's switch between them without having to do a bit of work :)

Perfect!

> What do you think about adapting your logging approach to have two loggers, one for console and another for a log file, each with different verbosity levels (including Off)?

Sure thing. I cannot imagine someone disabling the errors output, but I am in anyway! :-p . I will come with something for you

> How does your logging approach compare to what I've already done? Is the main benefit that it allows output to go to different places? Will it be easy to transfer over to your approach?

For the time being, it is based on a class with a streamer, which makes easier to call it. It has also named variables (MOORDYN_DBG_LEVEL and so on) to improve readability. In a close future it will automagically log the file, line and function where the log message was emitted.

sanguinariojoe:
OK, I already reworked the log handler:

 - Two simultaneous outputs, to the terminal (using cout or cerr) and the file
 - Two different threshold levels for the terminal and the file, by default MSG for the terminal, and DBG for the file
 - An option to completely disable the output
 - A helper class to be inherited by the objects that need to access the Log
 - Convenient macros to log at each level: LOGDBG, LOGMSG, LOGWRN, LOGERR
 - The LOGWRN and LOGERR macros automagically register the file, line and function where they are called
 - All this already works with the writeLog option in the input file

Thus now it is as easy as inheriting LogUser class and log messages:

```
LOGMSG << "A message" << endl;
LOGERR << "An error, something like MoorDyn.cpp:401 Init() will be automagically shown" << endl;
```

sanguinariojoe:
BTW, I rebased the branch to get your changesin `dev` integrated

mattEhall:
Finally following up on this @sanguinariojoe:

Thanks for supporting verbosity options for both console and text log file output - that sounds really convenient. 

I understand the value of your object-oriented approach to the API. I think many potential MoorDyn users are simple coders like me, but I'm sure there are also more advanced uses from more advanced coders like yourself that we should also support. Maybe we can still also keep the old API as the MoorDyn-for-Dummies option (which I will use).

sanguinariojoe:

PR #20, opened 2022-03-04
WIP: Optimizing the time step

sanguinariojoe:
I introduced a new option, dtMFactor (we should look for  better name), which is the times the critical time step can be considered. 

For the time being the critical time step is just considering the natural period of the lines (their segments natural period actually), and the velocity of the nodes. Conditions on bodies, rods and so on shall be imposed as well

The tests are working fine. Some of them are a bit faster, some are a bit slower. Tat demonstrates that you can get stable simulations using slightly larger time steps than the critical one, but not to much actually.

### DIFF
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -18,6 +18,8 @@
 #include "Waves.h"
 #include "QSlines.h" // the c++ version of quasi-static model Catenary
 
+#include <algorithm>
+
 using namespace std;
 
 // here is the new numbering scheme (N segments per line)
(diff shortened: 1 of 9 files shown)


mattEhall:
Interesting. Choosing the dt based on line node natural frequencies (I think that's what you meant to say) is something I've thought about, but I wasn't sure it was worth putting in code versus just giving users some kind of guidance for making the calculation themselves, since it's relatively easy. 

I'd really like to see a description of the theory for this (equations, etc.) before we get into the code of it. Can you let me know what you have in mind for "conditions on bodies, rods"? That's not something I've thought of before. Every new feature adds complexity so I wouldn't want to add this feature unless if works comprehensively, for every part of the system. That includes lines with bending stiffness...

We should also keep in mind that this may become obsolete if there is an adaptive time step method in the future.

sanguinariojoe:
> Choosing the dt based on line node natural frequencies (I think that's what you meant to say) is something I've thought about, but I wasn't sure it was worth putting in code versus just giving users some kind of guidance for making the calculation themselves, since it's relatively easy.

Well, right now it is indeed disable by default, because the default value of dtMFactor is ridiculously large. So it is up to the uses to go by the critical time step, or manually set the one he likes (or both)

> I'd really like to see a description of the theory for this (equations, etc.) before we get into the code of it.

I am adding the theory in the doxygen doc. e.g.

https://github.com/sanguinariojoe/MoorDyn/blob/courant/source/MoorDyn2.hpp#L295

For the time being, the very only _a priori_ knowledge used is the line nodes natural period. The damping should be added as well.

> Can you let me know what you have in mind for "conditions on bodies, rods"?

Nothing yet, I shall check them. But they will probably define natural periods or someting alike, and have characterisitic lengths and velocities. The idea is that MoorDyn will ask the critical time step to every single object in the system and select the minimum one.

> We should also keep in mind that this may become obsolete if there is an adaptive time step method in the future.

I already added that as well. For the time being it is just based on the velocity of each node, granting that the node does not move more than half the segment lenght in a single time step. The problem is that when you detect a large velocity, it is already too late. I need to store the accelerations as well, using `v + 0.5 * rd * rdd` as the characteristic velocity of the node, so in the end you need to grant both: `dt < 0.25 * l / rd` and `dt < sqrt(0.5 * l / rdd)`.

Something similar will probably apply to the other objects

sanguinariojoe:
There is a totally new way to handle the time integration in https://github.com/mattEhall/MoorDyn/pull/31

It would enjoy something like this in the future anyway

PR #22, opened 2022-03-16
Line::getStateDeriv() optimization

sanguinariojoe:
One of the main bottlenecks of the code is obviously the solution of linear systems, which is called for each single line node. For instance in test_bodies_and_rods:

callgrind out 19020

Line:getStateDeriv() takes a vast majority of the total computation (79.82%), and inside that function the heaviest load is in LuSolve calls (11.13%). Replacing the LU solver by a more straightforward approach, unfolded with inlines for further optimization, the load of that chunk of code is significantly reduced:

callgrind out 18136

The linear systems solving (called Solve3()) is taking now 6.99% of the total load in Line:getStateDeriv(), becoming comparable with some other unoptimized functions like unitvector() and dotProduct(). That allows to shrink the overall load of Line:getStateDeriv() to 78.86%

Let me know so I replace the solver everywhere, and continue optimizing some other helper functions (like unitvector() and dotProduct())

### DIFF
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -1083,7 +1083,7 @@ void Line::getStateDeriv(double* Xd, const double PARAM_UNUSED dt)
 				{
 					cout << "   Error: NaN value detected in bending force at Line " << number << " node " << i << endl;
 					cout << lstr[i-1]+lstr[i] << endl;
-					cout << sqrt( 0.5*(1 - dotProd( qs[i-1], qs[i] ) ) ) << endl;
+					cout << sqrt(0.5 * (1 - dotProd3(qs[i-1], qs[i]))) << endl;
 
 					cout << Bs[i-1][J] << endl;
 					cout << Bs[i  ][J] << endl;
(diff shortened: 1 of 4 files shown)


sanguinariojoe:
Hey @mattEhall !

I aggressively optimized the Line::getStateDeriv() function (and fixed a messed index), so the final profile looks like

callgrind out 23344

After the optimizations the function (which is called 435600 times!) is consuming the 70.81% of the total execution time, significantly less than the 79.82% we departed from. However, those percentages are pretty biased because of many factors.

If I move back to the dev branch, compile the release version and run the test I get the following execution time:

```
real    0m2,133s
user    0m2,120s
sys     0m0,003s
```

However, if I compile this optimized version I get the following execution time:

```
real    0m1,324s
user    0m1,314s
sys     0m0,007s
```

**1.6x faster!**

sanguinariojoe:
I tried splitting node_mass() in 2 different functions for the diagonal and non-diagonal terms, but the performance is actually the same.

So I really think this can be merged when it suits to you @mattEhall

mattEhall:
This is a great speed improvement! Wow! My only concern is changes to the code that will reduce its intuitive readability or make it appear very different from the Fortran version with a functional difference (this latter point mostly just applies to formatting changes). I will accept this pull request once the purely formatting changes are removed from it (so it is easy to see what functionality is actually changed vs not changed). 

Do you have a sense of the speed gain from eliminating the for loops? If it is small, I might want to keep some of the loops for easier interpretation and editing.

sanguinariojoe:

PR #26, opened 2022-04-26
Line optimization

sanguinariojoe:
This PR goes a bit further than https://github.com/mattEhall/MoorDyn/pull/22 , conveniently introducing [Eigen](https://eigen.tuxfamily.org/index.php?title=Main_Page) (the standard for vector & matrix related computations). Thus the code can be terribly simplified, transforming this:

```
for (int J=0; J<3; J++)
	vi[J] = U[i][J] - rd[i][J];
```
			
on this:

`const vec vi = U[i] - rd[i];`

On top of that, Eigen introduces many linear system solvers, which it is also simplifying the code and improving its performance and reliability.

And speaking of performance. The **test_bodies_and_rods** test took

```
real    0m2,133s
user    0m2,120s
sys     0m0,003s
```
to run. With the most aggressive options (which would slightly reduce the accuracy/precision, but are anyway passing the tests flawless) the execution time has became
```
real    0m0,898s
user    0m0,896s
sys     0m0,000s
```

**X2.38 times faster!**

-----

In this PR just the Line object is optimized (and probably I will not upgrade the other parts of the code until I extend https://github.com/mattEhall/MoorDyn/pull/19 to all the entities and get merged).

I should also upgrade the other compilers (All of them should still be working, but it would be nice to have the possibility to set optimization options as well)


sanguinariojoe:
I think this is ready to merge

mattEhall:
Finally following up on this @sanguinariojoe:

Really impressive speed boost! 

I guess I have to give in and rely on Eigen now, eh? It does seem very powerful...

My big concern is code readability (especially for less advanced coders like me). I do want MoorDyn's code to be understandable by as many people as possible. In your example above about how the code can be "terribly simplified," I think the original two lines is actually more readable than the simplified one line. The two lines makes it clear that the operation is a vector subtraction and it does not rely on a separate library that the reader would have to look into separately.

I'm not too worried about the above example, but a change I think is a larger impact to code readability is the new vec2array types of functions. I would prefer to avoid those, to keep the code simple to read. In other words, keeping something like
```
for (int I=0; I<3; I++)
	Xd[3*N-3 + 3*i-3 + I] = rd[i][I];
```
instead of
```
moordyn::vec2array(acc,   &Xd[            3 * i - 3]);
```
What do you think? You didn't answer my question in #22 about how much unrolling the for loops helps with computations speed, but I'm guessing it is pretty small (don't compilers often do this anyways?). So we could avoid the vec2array style functions and the speed would still be great, right?

My hope is that we can find a compromise between your many speed improvements and some more simple approaches that will keep the code easy to understand. Maybe something like this:
* We keep the awesome updates to use Eigen
* Can we keep the old for loops rather than the new vec2array style functions?
* With the namespace you added, is it possible to avoid having to write moordyn:: so much?
* I propose we skip the single precision option (and skip the 'real' typedef you added). Numerical problems often occur when using single precision, so it seems like more trouble than it's worth. Unless you know an important use case for single precision.

I noticed you have a new v2 branch that contains a lot of your new features combined. Because so much code has been changed (and sometimes unchanged), I've finally given up keeping track of the commit history. I don't see how it's possible for me to be selective. Instead I'm just looking at the end result and comparing it with older versions. So if it would be easier for us to forget these individual pull requests and just look directly at merging in your final combined version, that is fine with me.

sanguinariojoe:
Hey Matt! I will be KO for a while because i had a premature baby.

Anyway, some fast comments:

 - As you said, I am preparing a v2 branch where i am packing all the improvements in a quite large pull request.
 - I can provide both kinds of APIs with a bunch of inlines, so that will not be a problem. The new one will be probably more flexible though
 - It will be definitely based on Eigen. After porting all the code, those calls to vec2array and array2vec will be rather scarce. In fact, they will be mainly limited to the C <-> C++ interface. The main benefit of having those functions vs slicing and assigning, vout[i] = vin[i], is that you are letting know other people that a fundamental change/typing/casting is happening.
 - The moordyn:: can be almost always removed, since almost everything is in the moordyn namespace. I like it to be explicit though (it is indeed one of the Python zen statements)
 - The compilers are usually trying to unroll loops when they consider that appropriate. Unfortunately they usually do that really bad... You can look for some info around. On top of that, using loops you are breaking the SIMD operations, i.e. it is very bad for performance. We can further discuss this when i am submitting the large PR, and find a correct balance

mattEhall:
Congratulations Pepe, and I hope everyone is doing well!
Thanks for still providing those quick and informative responses. When you are back, I will be less of a road block because I now have some help for working on it from this end. So we can help more!
All the best.

sanguinariojoe:

PR #31, opened 2022-08-02
V2 version candidate!

sanguinariojoe:
Behold my v2 candidate!!! 100% done while holding my baby :-p

Let's briefly discuss the changes a bit... (There are many other things I just do not remember now, but probably they are commented in the README)

P.S. In the README there are also some tasks that we would probably address before releasing

# New API

## Separated C/C++ APIs

Now the C and the C++ APIs are totally separated. Thus, you can use MoorDyn from a C code without needing a C++ compiler at all. Along this line, the C++ API is hold in the *.hpp files, while the C one is hold in the *.h files.

The C++ API would still enjoy for a better design, while the C API is perfectly ready.

## C API

For backward compatibility a v1-like C API is retained, which can be accessed by including

`#include <MoorDyn.h>` 

However, a new v2 API has been designed, with much more potential. Such API is accessed in a different header

`#include <MoorDyn2.h>`

The most important (and controversial) change in the API is that every single object in the system can be instantiated now. This is well described in the documentation, but a chunk of code can be of use here:

```
#include <moordyn/MoorDyn2.h>

int main(int, char**)
{
    MoorDyn system = MoorDyn_Create("Mooring/lines.txt");
    MoorDynConnection conn = MoorDyn_GetConnection(system, 1);
    double x[3];
    MoorDyn_GetConnectPos(conn, x);
    return 0;
}
```

Effectively, instead of referring to the objects by their id, they are specifically extracted and used. That is making impossible to make typos messing up the indexes between objects (since they have different types). It is also slightly improving the performance, and it is allowing to move the C API methods relative to each object to its own header file (Connection.h, Rod.h, ...)

Bonus track, when using the C API within a C++ code, everything is automagically handled:

```
    auto system = MoorDyn_Create("Mooring/lines.txt");
    auto conn = MoorDyn_GetConnection(system, 1);
```

## Error handling

Now the C API functions are always returning either a NULL object instance, or a code error. On the other hand, the C++ API is throwing exceptions, which can be converted to C error codes by using the MOORDYN_CATCHER macro.

## Doxygen documentation

Both the C and the C++ APIs are now documented with Doxygen.

Bonus track:
Now the Doxygen documentation is retained when building the Sphynx one
 
# Eigen3 based

The C++ side now uses Eigen3 all over the place, which is fast and reliable. It is also reducing the code verbosity, which is always good.

# CMake

Now the project has a CMake configuration system. The old legacy makefiles are retained anyway

## Tests

The system to run unary tests is based on CMake now, and it works like a charm. It can be even integrated in a Github/Gitlab CI/CD to check the tests before any release, so no regressions happens.

## Wrappers

There are wrappers to use MoorDyn within Python and Fortran (The Fortran code is really a mirror of the C API). A Matlab mex file can be also added in the future 

## Windows installer

CMake can generate a Windows installer, which can be easily distributed in the Releases page.

# clang-formatter

A clang-formatter def file has been added, so the code follows the same formatting everywhere, which is also close to the most popular standards

# Documentation

I wrote some documentation regarding the installation and usage in both Windows and Linux. I also slightly improved the README, which now focus on this v2 version. I also fixed the version headers (which were missing some times, or were even GPLv3)


sanguinariojoe:
The CI is ready and running!

An example of the Releases automatically generated:
https://github.com/sanguinariojoe/MoorDyn/releases

It is also creating the wheels for the Python packages, i.e.

`pip install --index-url https://test.pypi.org/simple/ moordyn`

sanguinariojoe:
Matlab incoming! Stay tuned!

sanguinariojoe:
Here we go!

For some reason I do not understand, everything works fine, but the test is marked as failing because:

`Unrecognized function or variable 'x'.`

It runs flawless anyway... Strange... Matlab way of life...

Anyway, I am done with this! It is ready to merge from my side!

djanders19:
This seems great - particularly a fan of the ability to instantiate everything/treat the whole system in a more OO manner. Do any of the tests evaluate ability of the system to mimic real world data? I would be very curious to see that (and I'm sure it will be forthcoming in a publication once the merge is resolved). FWIW, I cloned the v2 branch for some of my own experimentation and observed some small numerical instabilities in my system that didn't seem to occur in the dev branch that I was previously working with - granted, I was working with time-steps that were larger than what's typically considered appropriate as I'm trying to see how much I can push performance while still maintaining a relatively workable degree of accuracy, but I did find it curious. Reducing the time-step slightly got rid of the instabilities at least qualitatively, but haven't had time to quantitatively compare to real data.

Also just out of pure curiosity - do you have numbers on how much faster this is running relative to the dev branch in your tests currently?

sanguinariojoe:
> Also just out of pure curiosity - do you have numbers on how much faster this is running relative to the dev branch in your tests currently?

You can get an idea [here](https://github.com/mattEhall/MoorDyn/pull/26). It is hard to give a final number though, it depends on the specific case, compiler, OS and hardware.

sanguinariojoe:
> Do any of the tests evaluate ability of the system to mimic real world data? I would be very curious to see that (and I'm sure it will be forthcoming in a publication once the merge is resolved).

Nope. If you take a look to the [readme](https://github.com/sanguinariojoe/MoorDyn/tree/v2#readme), there is an open task regarding tests. I really think we need to add some tests where we compare with analytic solutions or experimental data.

> FWIW, I cloned the v2 branch for some of my own experimentation and observed some small numerical instabilities in my system that didn't seem to occur in the dev branch that I was previously working with - granted, I was working with time-steps that were larger than what's typically considered appropriate as I'm trying to see how much I can push performance while still maintaining a relatively workable degree of accuracy, but I did find it curious

I made a ton of changes, and some of them corrected some stuff. On top of that, I probably messed up here and there. So if you detect glitches, and you can share the input files, that would help

By the way, in this candidate you can use some other time integration schemes, see the tScheme option at https://github.com/sanguinariojoe/MoorDyn/blob/v2/docs/usage.rst#options-1

djanders19:
Ah excellent! I'd also been experimenting with implementing other time integration schemes - looks like you beat me to the PR 😄. So there's also always the possibility that my own code lead to the error. I'll ask my mgmt about sharing inputs - some of it is derived from internal models so may not be publicly sharable. I'll also ask if we can share some experimental validation data with the project longer term.

My assumption is that the stability/accuracy of RK4, for instance, will be so much higher than the old RK2 that it is possible to push the timestep much higher for most systems and easily make up for the extra computational cost. I haven't encountered Adams-Bashforth before, but that also looks promising. Have you considered the performance of implicit schemes? My assumption is that they would be less performant given that mooring systems are generally not particularly stiff, but I've still been able to eke out better performance with implicit solvers in other libraries.

Overall I think this PR looks great - I'm focused on another project for work this week but am excited to start testing it with our workflow ASAP. Particularly once the output issue is fully resolved this could be a very useful system for our modeling.

sanguinariojoe:
In the [previous commit](https://github.com/mattEhall/MoorDyn/pull/31/commits/a0209073aa70f7daec222d144a4da64bcea2c32d) I added a pendulum test, which has an analytic solution.

Another interesting idea would be a "weightless" (same density than water) rope fixed to both end points, enforced by a constant current. That shall converge to a catenary shape

sanguinariojoe:
VTK support!

image

It is disable by default yet

PR #34, opened 2022-10-06
Bug fixes in code for dynamic currents

djanders19:
This is a relatively minor PR fixing some bugs in the dynamic currents code.

1. When reading inputs into UProfileZ, I got indexing errors with the old code. Updating to just push the currents to the back of the vector fixed that error.
2. When interpolating currents over the time dimension, the program keeps track of a variable `iti` that indicates the initial "upper index" of interpolation. However, this interpolation is performed at each depth, and if not reset in the inner loop winds up defaulting to the max time after running through all the times once. There may be a better way of fixing this by modifying inter_factor() in misc.h, but for now this appears to fix the bug.

Issue #28, opened 2022-07-06
Program fails to create Mooring system with large numbers of output files


djanders19:
With large numbers of output files (1022+ on my system) MoorDyn throws an error when creating .out files:
`Cannot create the output file './moordyn_inputs/test_with_10000_lines_Line1022.out
Error (-2) during the Mooring System creation:
Exception while reading the input file`

I believe this to be an issue with how files are handled. After outfiles are created, they appear to be kept open until the MoorDyn deconstructor is called. Most systems have a limit on the maximum number of open files at any given time, so I believe closing output files when they are not being actively written to or read from should solve the problem. I have tested closing the output files after they are created and this appears to solve the above error - however, it introduces more bugs as the downstream methods assume that all the outfiles are open.

There may not be many use-cases where this is an issue (I don't know how large people's models tend to get) so this refactor may not be worth the time/energy and presumed performance hit from opening and closing every output file every time the program wishes to write outputs. A more extensive, and possibly better, refactor would be to have all output saved to one larger file (presumably following Tidyverse principles or something similar) and just letting the user pull out individual Object outputs as part of their processing pipeline.

mattEhall:
This is a good point of discussion. I have rarely heard of scenarios that require more than 100 or 200 Lines -- even with some of our shared mooring array simulations, but I can imagine there will be important exceptions. For how to handle it, my thinking is to have a single file as you suggest. Along with that, have been hoping to add the capability for binary output files to save space and time. We should keep this in mind and discuss more as we get closer to finishing version 2!

sanguinariojoe:
The [v2 candidate](https://github.com/mattEhall/MoorDyn/pull/31) is already able to save binary files, which are BTW portable and platform agnostic.

It is though dangerous to save them in a double precission compiled MoorDyn (the standard) and load them in a single precission one (which is an option to boost the performance at the cost of accuracy).

I am planning to add VTK support for post-processing, but that will happens after  [the PR](https://github.com/mattEhall/MoorDyn/pull/31) is merged

djanders19:
So does the binary file update resolve the limited-line-number issue? I can see how it could reduce memory requirements quite a bit, but my understanding was that with large numbers of lines the program actually runs into an OS imposed limit on the number of files that can be open (as opposed to simply running out of memory). Could be totally wrong about that though - when I get some time I can try cloning that PR and running some 600-1000 line tests through to see if it still crashes the program.

[...]

PR #35, opened 2022-10-06
Single outfile

djanders19:
This PR addresses issue #28 by adding an option that allow the user to have all lines output to a single output file. The user can opt into this behavior by adding `1    outputMode` to the environment setup at the end of their input file. In this mode, all Line objects share a single shared_ptr to one output file. The resultant file is essentially just all the separated output files left-joined on Time. Do note that as it's currently setup, I've dropped the units line. In the future (if we continue to develop this feature) that'll need to be remedied. Also, this currently only works for Lines, but could be reasonably easily expanded to Rods and Bodies.

I'm also including the fixes to Dynamic Currents from #34 here, so we can close that and just merge this if it looks good.

Additionally, I included the RK4 implementation we had been using internally at Kelson before #31 was submitted with a bunch of new solvers. Once #31 is integrated that should be deleted, but for now I included it as it's easy to compile and use with the older system of building. If we'd rather _not_ deal with the bit of added complexity that could add to the ultimate future merge let me know and I will cherry pick those commits out.

### DIFF
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -162,77 +162,92 @@ void Line::initializeLine(double* X)
 		
 			// output time
 			*outfile << "Time" << "\t ";
+			string name = getLineName();
 			
-			// output positions?
+			// output positions?:w
+
 			//if (find(channels.begin(), channels.end(), "position") != channels.end())
 			if (channels.find("p") != string::npos)
 			{
 				for (int i=0; i<=N; i++)	//loop through nodes
 				{
-					*outfile << "Node" << i << "px \t Node" <<  i << "py \t Node" <<  i << "pz \t ";
+					*outfile << name << "Node" << i << "px \t" 
+					         << name << "Node" <<  i << "py \t"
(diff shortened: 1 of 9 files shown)


source/Line.cpp
```diff
@@ -1326,13 +1342,25 @@ void Line::Output(double time)
...
+				*outfile << "\n";
 		}
 		else cout << "Unable to write to output file " << endl;
 	}
 	return;
 };
 
+ofstream* Line::getOutputFilestream()
+{
+	return outfile;
```
sanguinariojoe, line 1355: We should really stop doing this... It is awful :-p

We must start using std::move

sanguinariojoe:
@mattEhall I am cherry picking the single-file output commits to submit another PR without merge conflicts

sanguinariojoe:
See https://github.com/mattEhall/MoorDyn/pull/37

sanguinariojoe:
This was cherry-picked. I suppose it can be closed

PR #37, opened 2022-10-15
Single outfile

sanguinariojoe:
Cherry-picked from https://github.com/mattEhall/MoorDyn/pull/35

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,7 @@ compile/SO/test*
 .eggs
 moordyn.egg-info
 tests/__pycache__
+
+# Visual studio stuff #
+#######################
+source/.vscode/settings.json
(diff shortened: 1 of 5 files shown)


sanguinariojoe:
@djanders19 , please, check this out... I did not even tested it

sanguinariojoe:
I rebased this PR to v2

@djanders19 , did you test it?

mattEhall:
@djanders19, @sanguinariojoe - am I right that this PR is where the nice option of a single output file lives, but it needs checking before being merged in?  

I've just merged in #51, where the many other PRs were nicely combined.

Are there any other outstanding items from #35 that we should think about, or was that all covered in #51?

sanguinariojoe:
Yeah, this was not tested. That's why I excluded it from #51

AFAIK all changes on #35 are included in both this PR and #51

PR #49, opened 2023-02-12
Fix https://github.com/mattEhall/MoorDyn/issues/47

sanguinariojoe:
Now Eigen and VTK will be required as dependencies if MoorDyn was compiled and installed with `-DEXTERNAL_EIGEN=ON` and `-DUSE_VTK=ON` flags respectively. Probably in the future we should create another target for the C wrappers (moordyn::moordync), so all this dependencies would affect just to the C++ API

I have also removed the "COMPONENTS" from the installations. They we useless nevertheless.


sanguinariojoe:
Closed in favor of https://github.com/mattEhall/MoorDyn/pull/51

PR #63, opened 2023-04-12
Fixes for rods

sanguinariojoe:
Just to let you know that I am working on fixing rods, so @mattEhall do not waste your time on them

### DIFF
--- a/docs/angles.svg
+++ b/docs/angles.svg
@@ -1,0 +1,203 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="97.131653mm"
+   height="104.65861mm"
+   viewBox="0 0 97.131653 104.65861"
+   version="1.1"
+   id="svg5"
+   xml:space="preserve"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   sodipodi:docname="angles.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"><sodipodi:namedview
+     id="namedview7"
+     pagecolor="#ffffff"
+     bordercolor="#000000"
+     borderopacity="0.25"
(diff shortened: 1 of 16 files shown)


sanguinariojoe:
Well, the rods seems to be working now. But there is an open question though...

Now the rods are featured, at least for the sake of integration, by the end point A and the orientation angles. However, the coupling is still done in terms of direction vector. That is obviously introducing errors of the order of the rotation velocity by the outer time step (the one used when you call `MoorDyn_Step()`)

I suppose the way to go is changing the rods coupling to work with the orientation angles as well, but any thought is welcomed @mattEhall and @djanders19

sanguinariojoe:
"errors of the order of the rotation velocity":

Let's figure an horizontal rod, q = [1, 0, 0], rotating at a constant speed, w = [0, 0, pi]. Let's say the inner time step is dtM = 0.001 (it does not matter too much) and the outer time step is dt = 1.0 (the one we are passing to https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/source/MoorDyn2.h#L191).

Then, x = q, and xd = [-2, 0, 0]. The q vector will become [1, 0, 0] for the first half of the outer time step, and [-1, 0, 0] for the second half.

That is a super-special case, of course. But in general the rotation will be represented with an error of order O(w * dt)

mattEhall:
The case you described involves combines a rapid rotation with a very large coupling time step, such that something rotates 180 deg over a single coupling time step. I'd consider that just a bad setup, because the coupling time step should always be small enough so that the dynamics at the coupling point can be approximated as linear between time steps. I.e. I'd ensure the coupling time step is small enough to keep rotations to under 5 degrees per step. Otherwise there's no way it would be reliable dynamic coupling.

Is that the only issue?

sanguinariojoe:
That's why I say the error is or order O(w * dt). If the outer step/coupling step by the rotation speed is small, the error will be small.

The point is the rods are internally handled by position and orientation angles now. Thus we can ask for angles instead of directions on the coupling, so there will be no error anymore. On the other hand, angles would be less intuitive.

So we just need to decide the way we want to go:

 - Receiving orientation angles (EulerXYZ) from the user
 - Changing the documentation to state that the outer time step should be small enough to grant small errors

mattEhall:
I think we should keep the coupling as before, using direction vectors for the rods.

The outer time step should always be small enough to grant small errors, whether in rotation or translation!

sanguinariojoe:
Ok! Then I am merging this!

PR #64, opened 2023-04-12
v2 Updates from Kelson

djanders19:
This PR covers a fair bit of work that we've done on the v2 codebase at Kelson over the last several months. To summarize these changes:
* Added support for a 4D current grid. This allows the user to specify a current-grid that varies in (x, y, z, t) space. A test is included, though the current-grid is too large to push to GH so we've included a python script to generate the current-grid for the test.
* First pass implementation of a 3D seafloor. This allows the user to specify as input a 2D-grid of z-values specifying the bathymetry of a site. The seafloor is implemented as it's own class - model objects can look up the depth for an arbitrary X/Y point by calling the getDepthAt(real, real) method of this class. Currently, all the physics force calculations remain the same - the only thing that has been changed is how objects determine the seafloor height at their location when checking for contact. For example, an object in contact with a sloped seafloor won't "slide" down the slope due to gravity. An example 3D seafloor input file is seafloor_profile_3d.txt and a test is included in tests/Mooring/seafloor.txt and tests/seafloor.cpp
* Partial submergence in still water (i.e. no waves condition) is implemented for lines. This is _not_ currently implemented for the waves condition, but that feature will be coming in a future PR.
* Bugfixes for interpolation, particularly interp4. As an example of the issue we were encountering, this is what was happening when we interpolated wave velocities (they now interpolate smoothly):
image
* Bugfixes in line logging
* We have started a refactor of the readInFile() function in MoorDyn2. Mostly, we've just summarized some repetitive code in functions and started to pull out the code for reading particular objects in into their own functions as well. Ultimately, we're trying to do this for all long functions to reduce code reuse and generally make the code more approachable. This is very much a WIP, but we want to keep things as synced up as possible.
* We've added a fair bit to the .rst docs, describing some of the model objects and some of the new features above. Ultimately, we want to make this code presentable to engineers who have maybe not worked in an OOP paradigm before.


docs/starting.rst
```diff
@@ -172,7 +181,7 @@ Press "Configure" once again, and then "Generate". Now you can close CMake.
 
 Now, since we are installing Eigen in C:\Program Files (x86)\Eigen3, we need
 to execute a Command Prompt with administrative rights.
-To this end, look for "cmd" in the Windows Init menu and do right click on
+Search for for "cmd" in the Windows Init menu and right click on
```
sanguinariojoe, line 184: "for for"

source/Body.cpp
```diff
@@ -336,16 +339,18 @@ Body::initiateStep(vec6 r, vec6 rd)
...
 	}
 	if (type == FIXED) // if the ground body, set the BCs to stationary
 	{
 		r_ves = vec6::Zero();
 		rd_ves = vec6::Zero();
-		return;
 	}
-	LOGERR << "The body is not a coupled/fixed one" << endl;
-	throw moordyn::invalid_value_error("Invalid body type");
+	else
```
sanguinariojoe, line 348: It is a common practice to use the returns to shortcut the function execution rather than an else. It does not matter much anyway

source/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ set(MOORDYN_HEADERS
     MoorDyn.h
     Rod.h
     Waves.h
+    Seafloor.hpp
```
sanguinariojoe, line 39: I suppose the answer is no... But would it be convenient to add an .h file with C functions, so the users can at least read the bathymetry? (Maybe even edit it)

source/Connection.cpp
```diff
@@ -145,7 +146,11 @@ Connection::initialize()
 		pos = r;
 		vel = rd;
 
-		if (-env->WtrDpth > r[2]) {
+		real waterDepth = -env->WtrDpth;
```
sanguinariojoe, line 149: const real waterDepth?

source/Connection.cpp
```diff
@@ -212,10 +217,11 @@ Connection::GetConnectionOutput(OutChanProps outChan)
 }
 
 void
-Connection::setEnv(EnvCond* env_in, moordyn::Waves* waves_in)
+Connection::setEnv(EnvCond* env_in, moordyn::Waves* waves_in, std::shared_ptr<moordyn::Seafloor> seafloor_in)
```
sanguinariojoe, line 220: It is a bit strange that we are passing by pointers and just seafloor is using std::shared_ptr. Probably we shall consider moving everything to shared_ptrs

source/Line.cpp
```diff
@@ -473,7 +473,8 @@ Line::initialize()
 
 	// if conditions are ideal, try to calculate initial line profile using
 	// catenary routine (from FAST v.7)
-	if (-r[0][2] == env->WtrDpth) {
+	real waterDepth = getWaterDepth(r[0][0], r[0][1]);
```
sanguinariojoe, line 476: const real waterDepth?

source/Line.hpp
```diff
@@ -95,12 +97,39 @@ class Line : public io::IO
...
+		}
+		else {
+			return -env->WtrDpth;
+		}
+	}
 
+	/** @brief A single value representing the average water depth
+	 * 
+	*/
+	real avgWaterDepth() {
```
sanguinariojoe, line 118: This can be clearle an inline function

source/Line.hpp
```diff
@@ -95,12 +97,39 @@ class Line : public io::IO
...
+		if (seafloor) {
```
sanguinariojoe, line 119: You can use an inline if:

`return seafloor ? seafloor->getAverageDepth() : -env->WtrDpth;`

Up to you!

source/Misc.hpp
```diff
@@ -861,14 +872,14 @@ interp4(const vector<vector<vector<vector<T>>>>& values,
...
 
-	T c000 = values[i0][j0][k0][w0] * fw + values[i0][j0][k0][w] * (1. - fw);
-	T c001 = values[i0][j0][k][w0] * fw + values[i0][j0][k][w] * (1. - fw);
-	T c010 = values[i0][j][k0][w0] * fw + values[i0][j][k0][w] * (1. - fw);
-	T c011 = values[i0][j][k][w0] * fw + values[i0][j][k][w] * (1. - fw);
-	T c100 = values[i][j0][k0][w0] * fw + values[i0][j0][k0][w] * (1. - fw);
-	T c101 = values[i][j0][k][w0] * fw + values[i0][j0][k][w] * (1. - fw);
-	T c110 = values[i][j][k0][w0] * fw + values[i0][j][k0][w] * (1. - fw);
-	T c111 = values[i][j][k][w0] * fw + values[i0][j0][k0][w] * (1. - fw);
+	T c000 = values[i0][j0][k0][w0] * (1. - fw) + values[i0][j0][k0][w] * fw;
```
sanguinariojoe, line 875: Can you please check if this error is affecting also other interpolations? It is harder for me to do that from my cellphone

source/MoorDyn2.hpp
```diff
@@ -432,13 +465,17 @@ class MoorDyn : public io::IO
 
 	/// The time integration scheme
 	TimeScheme* _t_integrator;
+	/// The time integration scheme name
```
sanguinariojoe, line 468: The time integrators have an inline method to retrieve the name, so this is quite redundant. If you need this while parsing the input file, I think it should be rather a local variable

source/Rod.hpp
```diff
@@ -223,6 +226,20 @@ class Rod : public io::IO
 	 * @}
 	 */
 
+	/** @brief Finds the depth of the water at some (x, y) point. Either using env->WtrDpth or the 3D seafloor if available
+	 * @param x x coordinate
+	 * @param y y coordinate
+	 * @return A negative number representing the sea floor depth at the given location
+	*/
+	real getWaterDepth(real x, real y) {
```
sanguinariojoe, line 234: inline function

source/Rod.hpp
```diff
@@ -223,6 +226,20 @@ class Rod : public io::IO
...
+		if (seafloor) {
```
sanguinariojoe, line 235: You can do an inline if, see above

sanguinariojoe (changes requested):
All looks nice! The compilation failed though (see https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/4682752558/jobs/8304525686?pr=64)

Also, please remember to:

 - Add your organization and any other contributor at tge end of https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/README.md
 - Increase the version on https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/CMakeLists.txt#L4

I will further check when i have a computer (i am taking some days off)

source/Connection.cpp
djanders19: Yeah, I also think it would make sense to move to shared_ptrs. I will discuss with Alex (our new guy) and see if he sees any reason not to - I think the raw pointers are just a holdover from the initial C->C++ conversion. 

Really, what we need is to spend a week or two just refactoring things like this.

source/CMakeLists.txt
djanders19: As in, include the bathymetry features in the C API? I don't _think_ that should be too inconvenient. Will have to take a look at the conventions for setting up those headers though, as I've pretty much only used the c++ updated version.

source/Line.cpp
djanders19: Done.

source/Body.cpp
djanders19: Do you mean return in the if blocks so as not to execute full conditional? If so, done.

docs/starting.rst
djanders19: Fixed, thank you!

source/Line.hpp
djanders19: I guess this gets to the great ternary vs. classic conditional readability debate. I also prefer the oneliner for things like this so I'll make that update.

source/Line.hpp
djanders19: Done. Does explicitly declaring inline have much of an effect these days? [I've been lead to believe it does not](https://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method).

source/Rod.hpp
djanders19: Done.

source/MoorDyn2.hpp
djanders19: Agreed. I think the reason for this originally being implemented this way is that in the current setup, timeschemes are created using a create_time_scheme() function in the MoorDyn2::readInFile() member function as opposed to the MoorDyn2::readOptionsLine() member function, so we can just change the time integrator to be setup there and only do the default setup with RK2 if that options line isn't specified.

More generally, is there a reason to keep the time integrator setup in the create_time_scheme() function as opposed to the constructor (or maybe a builder)?

source/Misc.hpp
AlexWKinley: For what I can see it looks like all the other interpolations are correct.

djanders19:
I believe we've now addressed all the above points. All tests are passing on our end. To summarize:
- Fixed Linux compilation errors resulting from case-sensitivity in includes.
- Added basic C API for Seafloor
- Updates to use smart pointers to reference environment data structures
- Got rid of redundant t_integrator_name member variable
- Various style updates as outlined in conversations above.

source/Connection.cpp
djanders19: @AlexWKinley addressed this in ddd95a8144288ce8e084aa28037ae6ffd561f842

source/CMakeLists.txt
djanders19: Done.

source/MoorDyn2.hpp
sanguinariojoe: Well, `create_time_scheme()` is almost a map which gets a name and returns a time scheme. However, there are some special cases, like the implicit schemes. That's why I made it a function. On the other hand I think it is quite appropriate to hold it on Time.hpp and Time.cpp.

Anyway, if you find it a better place or implementation in general, I obviously do not oppose

sanguinariojoe (approved):

source/Body.hpp
```diff
@@ -49,24 +54,29 @@ using namespace std;
 namespace moordyn {
 
 class Waves;
+typedef std::shared_ptr<Waves> WavesRef;
```
sanguinariojoe, line 57: This is a bit ugly... I really like predeclarations, but here it is way better to move the Waves.hpp include here, so we do not need to redefine WavesRef.

source/Connection.hpp
```diff
@@ -49,16 +50,17 @@ namespace moordyn {
 
 class Line;
 class Waves;
+typedef std::shared_ptr<Waves> WavesRef;
```
sanguinariojoe, line 53: Same... Include Waves.hpp

source/Rod.hpp
```diff
@@ -49,6 +50,7 @@ using namespace std;
 namespace moordyn {
 
 class Waves;
+typedef std::shared_ptr<Waves> WavesRef;
```
sanguinariojoe, line 53: Same... Include Waves.hpp

sanguinariojoe (approved):
I added some comments, but actually i will merge right away, and fix them myself. No point on bothering you any longer

Thanbks!

PR #71, opened 2023-05-12
Kelson Waves Features and Refactor

AlexWKinley:
This pull request contains a number of new features as well as major refactor of the wave and current code in MoorDyn.

This is definitely a big pull request containing a lot of work and changes.
We're happy to discuss the rational for why we did things certain ways, or if there are things that should be changed.
Additionally this could be a place to discuss any potential future waves features and how they may be impacted by this refactor. 

We have not yet done a version bump in part due to uncertainty about if this should be considered a major or minor version change.

## Major New Features

The new features are mainly related to waves and currents.

### Wheeler stretching for wave kinematics

The wave kinematic calculations make the assumption of a flat seafloor and a flat water surface at z = 0.
This assumption is not locally true given wave height changes to the water surface as well as the potential 3D seafloor.
Wheeler stretching has the effect of linearly stretching the water column between the water surface and seafloor.
This prevents points within a wave at z > 0 from having unrealistically high wave forces, and a point on a raised part of the seafloor will not have vertical wave forces.
Currently this stretching happens for all of the internally calculated wave modes, it does not get applied to user supplied external waves.

### A New Wave Mode: WaveKin = 7 WAVE_SUM_COMPONENTS_NODE

We have implemented a new wave mode that takes the same `wave_frequencies.txt` file that the FFT Wave grid mode does, but instead of precomputing a wave grid, it directly calculates the sum of the water kinematics from each component wave at every node. 
The advantage of this method is that it is entirely accurate for the moment it is calculated for, and doesn't have the memory usage of the wave grid methods. It also allows for any number of input frequencies with any frequency spacing without needing to interpolate them to be evenly spaced. You can even have multiple components with the same frequency, which is useful if they have different directions (see the wave direction enhancement).
The disadvantage is that even for a small number of wave frequencies, it is slower to calculate the sum of the component waves, than to do the wave grid lookup and interpolation. 

### Line Partial Submergence with Waves

Line segments now will compute their submergence when any of the wave modes are being used.
The submergence is computed using a flat plane with a height equal to the average of the surface height at segments two nodes.
To be a reasonable approximation this requires that lines near the surface be short enough that the water surface is roughly horizontal across the area of the segment. 

### Wave Component Direction in `wave_frequencies.txt`

The two wave modes that use wave spectra now support per-component wave direction. 
The input format has been expanded to support reading a 4th value on each line that is the wave angle in radians.
If there is no 4th value specified on a line it defaults to 0 degrees, which was the previous MoorDyn behavior. 
For the Wave FFT Grid mode, this is probably mostly useful for specifying the same angle for all components.
For WAVE_SUM_COMPONENT_NODE, this allows you to specify multiple wave spectra each in different directions in one file. 

An example of the changed file format: 
```
0.0 0.0 0.0
0.35 1.0 0.0 0.0
0.4 0.0 1.0 1.57
0.6 0.0 0.0
```

This specifies two waves each with amplitude of 1 (but with a pi/2 phase offset), that also have a pi/2 angular offset so that one travels in the +x direction, and one travels in the +y direction.

### Wave and Current Documentation

A new page has been added to the documentation that covers all of the wave and current modes including their input file formats and general behavior.
Additionally the code for the waves refactor has been relatively thoroughly documented, and the doxygen config has been modified to make sure the subfolders get included in the documentation.
Looking through the doxygen documentation is a decent way to get a sense of how things work after the refactor especially due to the inheritance and collaboration diagrams that doxygen makes.

### External Wave Kinematics for Bodies, Rods, and Connections

When using external waves, bodies, rods, and connections will have their locations returned by `MoorDyn_ExternalWaveKinGetCoordinates`. 
They also contribute to the number of points returned by `MoorDyn_ExternalWaveKinInit` and `MoorDyn_ExternalWaveKinGetN`. 
The values returned for their points will then be used by those structures for any water force calculations they do.

We have also added the ability to use one of the current modes and external waves and have those be added together. 
This is likely not super practically useful, but could simplify certain use cases.

### Basic Benchmarks using Google Benchmark

We have integrated the Google benchmark library into MoorDyn and made some initial benchmarks primarily looking at the performance of different wave modes. 
This makes it easier to track MoorDyn's performance over time or with different settings.

### Added the Option to not Combine Wave Grids and Current Grids

Previously, when you were using a wave grid mode, and a current grid mode, MoorDyn would interpolate and add the current kinematics onto the wave grid. 
This remains the default behavior, but the new `UnifyCurrentGrid` option can be set to `0` to disable this behavior.
Instead two grids will be built and saved separately.
Then when calculating the water kinematics they will individually be interpolated and the current velocities and accelerations will be summed.
This can be useful if you have wave and current grids with significantly different locations or resolutions.
The practical benefit of this is not particularly big, but in cases where the wave grid cover a shorter amount of time than the current grid, this prevents the currents from being cut off early and looping in an unintended way. 

### More Tests

A number of new tests, a number of which are internally commented out, have been added.
Some of these tests mainly serve to demonstrate usage and options. 
They are also potentially useful to someone doing future waves work to ensure they don't break existing behavior.

## Waves Refactor

A major component of the amount of code changes is the fact that we have nearly entirely refactored the wave and current code.

This refactor centralized and unified a lot of the waves code.
Structures no longer have to keep track of their own wave kinematics, and do not have to care about the wave mode.
Additionally it is easier to add new wave or current modes in the future.

### High Level Description of the Changes

The Waves class now is the central controller and manager of wave and current kinematics. 
The Waves class knows about all of the structures (lines, connections, etc) in the simulation, and can get all of the node positions for the structures. 
All of the calculation of wave or current grids, and interpolating to calculate the kinematics of waves or currents at a specific point have been moved out of the Waves class and into their own functions or classes. 
Two new abstract classes `AbstractWaveKin` and `AbstractCurrentKin` have been created that each specify a simple "interface" that a class can inherit and implement to denote that it can provide wave or current information.
This means that the Waves class just needs to keep an AbstractWaveKin pointer and an AbstractCurrentKin pointer, and does not need to be concerned with how exactly the wave or current calculations are done. 

The existing wave and current modes have all been converted to this method. 
Specifically the `WaveGrid` and `CurrentGrid` classes (found in Waves.hpp/Waves.cpp) can be filled with data during the setup phase depending on the options and input data. 
This new architecture also means that the new wave mode just needs to be handled inside of the Waves code, and things like lines and connections have no knowledge of the new wave mode, or any of the wave modes anymore. 

The changes to the code that enabled this are significant, and as part of this we have moved existing code around to prevent files from getting unreasonably long and help keep things organized. 
If there are specific questions or concerns about this refactor I am more than happy to have a more detailed discussion about this work. 

## Bug Fixes and Misc Changes

A number of bug were found and fixed in the process of writing and testing this refactor. 
These include:

- An off by one error in `interp_factor` that would cause issues when the coordinate array was of length 2
- Wave FFT Grid now correctly calculates and sets its own dtWave depending on the frequency spacing of the input
- Some windows specific issues related to cmake and building the docs

Some miscellaneous changes were made either for convenience or to simplify future work.
These include:

- Enabled a cmake option that exports all symbols in the dll on windows
- If the cmake version supports it, cmake now precompiles the misc.hpp header in an attempt to speed up compilation, this didn't seem to make a huge change but is maybe something to tweak for the future
- Added some utilities in MoorDyn in and in the tests for working with writing VTK files and vtk .series files that can be imported by Paraview

## Breaking Changes and Changes to Default Behavior

Most of this work was done with the intent of maintaining backwards compatibility but for various reasons this PR contains a number of changes to behavior or otherwise breaks backwards compatibility. 

### Saving and Loading

Because the actual water kinematics data got moved out of lines, connections, etc, they no longer get saved and loaded. 
This means that waves and currents do not get saved or loaded. 
Because loading a model first requires you to call setup, this generally does not affect anything.
We have not deeply investigated if there is a situation or a use case that this does effect. 
It would not be impossible to reimplement saving and loading of waves and current if it was deemed necessary.
Even in the case where we reimplement saving and loading of waves, it would likely still mean that you could not load models saved with previous versions of MoorDyn.

### Extrapolating/Interpolating External Wave Kinematics

Previous versions of MoorDyn would do acceleration based extrapolation of external wave kinematics. 
Additionally it was possible to define waves at a point in the future to be interpolated to (at least I'm pretty sure if I'm reading the code).

This behavior has not been carried over to the refactor. 
Instead, water velocities and accelerations are treated as constant during that time interval.
This change is noted in the new wave documentation.
The new waves architecture does not prevent reimplementing the extrapolating/interpolating behavior, but 
we were uncertain on if that behavior was always preferable to constant velocities and accelerations. 

This change is definitely up for discussion if there is a need for the extrapolation or if there is a desire for the external wave kinematics to give the user more possibilities and options.


CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 endif()
 if((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME) AND BUILD_TESTING)
     add_subdirectory(tests)
+    add_subdirectory(bench)
```
sanguinariojoe, line 136: Do we want to always run the benchmarks when testing is enabled?

The benchmarks require third-parties, and internet connection. Having the possibility to disable them would be beneficial, am I right?

sanguinariojoe:
Guys, the building failed. Look at https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/4959833713/jobs/8947549748

The folder id called `Waves`, not `waves`

source/Body.cpp
```diff
@@ -345,9 +345,11 @@ Body::initiateStep(vec6 r, vec6 rd)
 		r_ves = vec6::Zero();
 		rd_ves = vec6::Zero();
 		return;
+	} else {
```
sanguinariojoe, line 348: Why is this needed? The returns are already short-cutting the function execution

Indeed this is wrong, since the else only affects to [the second if](https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/source/Body.cpp#L343)

source/Body.cpp
```diff
@@ -450,7 +451,7 @@ Body::doRHS()
 	// for rotation, this is just the negative of the body's rotation for now
 	// (not allowing flow rotation)
 	vec6 vi = -v6;
-	vi(Eigen::seqN(0, 3)) += U;
+	vi(Eigen::seqN(0, 3)) += U[0];
```
sanguinariojoe, line 454: Is this not wrong?

[U is already a vector](https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/source/Body.hpp#L144). `vi(Eigen::seqN(0, 3)) += U;` means add `U` to the first 3 components of `vi`. AFAIK that was right

source/CMakeLists.txt
```diff
@@ -65,7 +79,10 @@ set_property(TARGET moordyn PROPERTY
 set_property(TARGET moordyn APPEND PROPERTY
   COMPATIBLE_INTERFACE_STRING moordyn_MAJOR_VERSION
 )
-
+if(${CMAKE_VERSION} VERSION_GREATER "3.15") 
+  message("Using Precompiled Headers!")
+  target_precompile_headers(moordyn PRIVATE Misc.hpp)
```
sanguinariojoe, line 84: I actually never used this... Can we expect issues around inlines and so on? Do not think so, right?

sanguinariojoe:
I only had time to review until Connection.cpp... Sorry!

Anyway, please fix the typo so I can run the CI/CD and some other incidences I found. I will continue with the review after that, OK?

TY guys! Good work here!

source/Body.cpp
AlexWKinley: As part of the waves refactor, structures now get the water kinematics from the waves class, and don't store it internally. In this case, a few lines above Body gets its wave kinematics with
```c++
auto [zeta, U, Ud] = waves->getWaveKinBody(bodyId);
// U is of type const std::vector<vec3>&, 
// but will have a size of 1 because a body only needs water kinematics at a single point
```
For simplicity reasons, all of the Wave::getWaveKin**** functions (for lines, connections, rods, bodies) return the water kinematics in the form of a vector for each "node". This includes for bodies and connections, even though they only have a single node. Doing it this way just reduces code duplication with minor variations inside of the Waves class. Because of this, bodies and connections do need to manually get the one element in the vector.

source/Body.cpp
AlexWKinley: You are right, I think this must have been an accidental change on my part. I've returned it to the way it was previously.

source/CMakeLists.txt
AlexWKinley: My impression from reading about the feature is that it shouldn't behave any differently besides potentially increasing compile times when you change `Misc.hpp` and hopefully decreasing compile times when you change things that include `Misc.hpp`. But I don't have much concrete evidence on its effect on performance especially on compilers other than MSVC, but from what I've experienced it doesn't seem to hurt inlining or performance in other ways.

CMakeLists.txt
AlexWKinley: Considering the need for a network connection I do agree. I've added a cmake variable to control whether the benchmarks get built, and have them turned them off by default.

AlexWKinley:
I've hopefully fixed the first batch of low hanging fruit. Certainly take your time with the review, it's a lot of code and a lot of changes.

source/Body.cpp
sanguinariojoe: Right, [you are dropping indeed U from the header](https://github.com/FloatingArrayDesign/MoorDyn/pull/71/files#diff-a76fe5fd1cca7a26ba246ab4d1461719534ccb31d1305dd71cd517f7f3ffc064L144)

source/CMakeLists.txt
sanguinariojoe: OK, I suppose it is safe to add this. We will always have time to revert it

source/Util/VTK_Util.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @brief Returns the number of rows in an Eigen::Matrix that has 1 column
+ *
+ * @tparam T
+ * @tparam N
+ * @param mat
+ * @return constexpr unsigned int
+ */
+template<typename T, unsigned int N>
```
sanguinariojoe, line 27: The [compilation failed again](https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/5003644216/jobs/8982274314). Apparently on Linux you need the specialization

```
template<typename T, int N>
constexpr unsigned int
get_vec_size(const Eigen::Matrix<T, N, 1>& mat)
{
	return N;
}
```

I can even remove the `unsigned int` one. Probably is best to keep both

sanguinariojoe: Oh, very true! Eigen has the enigmatic `Eigen::VectorXd`...

bench/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+function (add_bench bench_name src)
+    add_executable(${bench_name} ${src})
+    add_dependencies(${bench_name} Bench_MooringConfigFiles)
+    target_link_libraries(${bench_name} benchmark::benchmark)
+    target_link_libraries(${bench_name} moordyn)
+    target_include_directories(${bench_name} PUBLIC ${CMAKE_BINARY_DIR}/)
+endfunction()
+
+add_bench(WaveBenchmark WaveBench.cpp)
+add_bench(MoordynBenchmark MDBench.cpp)
```
sanguinariojoe, line 24: How do I run the benchmarks? Shall I run the executables manually? If that is the case, maybe we can [create a custom target named run_benchmarks](https://cmake.org/cmake/help/latest/command/add_custom_target.html), depending on both WaveBenchmark and MoordynBenchmark

What do you think?

tests/CMakeLists.txt
```diff
@@ -16,18 +16,56 @@ if(WIN32)
...
+    time_schemes
+    io
+    rods
+    bodies_and_rods
+    quasi_static_chain
+    wavekin_4
+    tripod
+    seafloor
+)
+
```
sanguinariojoe, line 33: Move

```
set(CPP_TESTS 
    wavekin
    wavekin_7
)
```
here, that would help later

tests/CMakeLists.txt
```diff
@@ -16,18 +16,56 @@ if(WIN32)
...
+function(make_cpp_executable test_name)
+    make_executable(${test_name})
+    target_include_directories(${test_name} PUBLIC ${CMAKE_BINARY_DIR}/)
+endfunction()
+
 if(USE_VTK)
     set(TESTS "${TESTS};vtk")
+
+    make_cpp_executable(wave_surface)
```
sanguinariojoe, line 48: Replace this by `set(CPP_TESTS "${CPP_TESTS};wave_surface")`. That way the test will be automagically compiled and added to the tests batch (right now it is ignored)

source/Waves/WaveGrid.cpp
```diff
@@ -0,0 +1,897 @@
...
+	}
+
+	vector<real> UProfileZ;
+	vector<real> UProfileT;
+	vector<vector<real>> UProfileUx;
+	vector<vector<real>> UProfileUy;
+	vector<vector<real>> UProfileUz;
+
+	// this is the depths row
+	vector<string> entries = moordyn::str::split(lines[4]);
```
sanguinariojoe, line 662: [Now a test failed](https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/5014577378/jobs/9004047835?pr=71). The problem are the whitespaces at the tail of tests/Mooring/wavekin_3/current_profile_dynamic.txt. Probably we should add and use [a strip method](https://stackoverflow.com/questions/216823/how-to-trim-an-stdstring)

tests/CMakeLists.txt
AlexWKinley: In this case, this was an intentional decision. `wave_surface` doesn't do a lot of useful correctness testing, but does serve as a reference for saving wave elevation data to a vtk format. As well as being a way of testing line interaction with the wave surface as part of a development workflow. For this reason, and the fact that it does a lot of IO, I figured it would be best as an executable that gets created, but not as part of the normal ctest suite. If there is a different way you want to accomplish that sort of thing, I'm certainly open to suggestions.

AlexWKinley:
I hopefully solved the line ending problem. It seems like the main source of issues was windows line endings which would create an extra `\r` entry after calling split on the line, if there was also trailing spaces. I solved this with an rtrim function similar to the one you linked, and went through and changed hopefully all the instances of reading the lines of a file to either use rtrim, or call a function that does. 

This should also, hopefully, get the CI to run successfully. I don't have a Linux machine to test on, but running locally through docker with [act](https://github.com/nektos/act) seems to work.

bench/CMakeLists.txt
AlexWKinley: Easy enough, done.

sanguinariojoe:
Merged, TY!

PR #75, opened 2023-06-05
Minor bug fixes from Ryan

RyanDavies19:
Reading output options fix, rk2 fix, catenary line profile initialization fix, added simulation time to connection object for debugging purposes, and tangential Froude-Krylov force calculation fix.

### DIFF
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -522,13 +522,8 @@ Line::initialize()
 		}
 	}
 	else {
-		for (unsigned int i = 1; i < N; i++) {
-			r[i] = r[0] + dir * (i / (real)N);
-		}
-		else {
-			LOGWRN << "Catenary initial profile failed for Line " << number
-			       << endl;
-		}
+		LOGDBG << "Vertical initial profile for Line " << number
+		       << endl;
 	}
 	
 
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -505,12 +505,6 @@ Line::initialize()
 			if (abs(Zl[N] - ZF) > Tol) {
 				LOGWRN << "Wrong catenary initial profile for Line, intializing as linear " << number
 				       << endl;
-        // From above:
-        // initialize line node positions as distributed linearly between the
-        // endpoints
-        for (unsigned int i = 1; i < N; i++) {
-          r[i] = r[0] + dir * (i / (real)N);
-        }
 			}
 			else {
 				// the catenary solve is successful, update the node positions
@@ -521,14 +515,8 @@ Line::initialize()
 					r[i] = r[0] + l;
 				}
 			}
+		}
 		else {
-			// From above:
-			// initialize line node positions as distributed linearly between the
-			// endpoints
-			for (unsigned int i = 1; i < N; i++) {
-				r[i] = r[0] + dir * (i / (real)N);
-			}
-
 			LOGWRN << "Catenary initial profile failed for Line " << number
 					<< endl;
 		}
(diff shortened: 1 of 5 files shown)


source/Time.cpp
```diff
@@ -198,9 +198,8 @@ RK2Scheme::Step(real& dt)
 	r[1] = r[0] + rd[0] * (0.5 * dt);
 	Update(0.5 * dt, 1);
 	// And so we can compute the new derivative and apply it
-	CalcStateDeriv(0);
-	r[0] = r[0] + rd[0] * dt;
-
+	CalcStateDeriv(1);
```
sanguinariojoe, line 201: This change is actually not needed. In RK2 you can overwrite rd[0] since it is not needed anymore. I do not know if it is better for readability, but in my opinion it is better as it was before. I am open to change my mind though

source/Time.hpp
```diff
@@ -751,7 +751,7 @@ class HeunScheme : public TimeSchemeBase<1, 2>
  *
  * This was the traditionally applied time scheme in MoorDyn v1
  */
-class RK2Scheme : public TimeSchemeBase<2, 1>
+class RK2Scheme : public TimeSchemeBase<2, 2>
```
sanguinariojoe, line 754: This should be reverted if the Time.cpp change is eventually reverted

source/Time.cpp
RyanDavies19: Right, I didn't realize that. From a readability perspective this seemed like it was the Euler method to me which is why I changed it. Does it make a big difference in efficiency either way?

sanguinariojoe: Totally uncritical. I think it can be read just fine, but the fact that you got twisted might be a signal of pitfall. So you decide. Just let me know

RyanDavies19: I think we should change it. Both Matt and I were confused by that.

sanguinariojoe: Perfect then

source/Line.cpp
```diff
@@ -455,64 +455,77 @@ Line::initialize()
 
 	// if conditions are ideal, try to calculate initial line profile using
 	// catenary routine (from FAST v.7)
-	real waterDepth = getWaterDepth(r[0][0], r[0][1]);
-	if (r[0][2] == waterDepth) {
-		real XF = dir(Eigen::seqN(0, 2)).norm(); // horizontal spread
+
+	real XF = dir(Eigen::seqN(0, 2)).norm(); // horizontal spread
```
sanguinariojoe, line 459: These changes seem to break the pendulum test:

https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/5179676881/jobs/9420343210?pr=75

sanguinariojoe (changes requested):
Pendulum test has no reason to fail. Apparently it is ruined when trying to converge during the initialization, so I suppose it is related with the changes on Line::initialize().

The time_schemes test is also failing, but apparently you can solve that just decreasing [the time step for Backwards Euler](https://github.com/RyanDavies19/MoorDyn_ryan/blob/v2/tests/time_schemes.cpp#L75) to 1.7E-4. No big deal actually

RyanDavies19:
Thanks for that rebase. I am not seeing the pendulum test case causing any issues in the Fast version of MoorDyn, so I'm not sure if the bug is there too.

sanguinariojoe:
> Thanks for that rebase. I am not seeing the pendulum test case causing any issues in the Fast version of MoorDyn, so I'm not sure if the bug is there too.

Then something is different on https://github.com/FloatingArrayDesign/MoorDyn/blob/v2/source/QSlines.hpp . I think we can merge this and submit an issue (the tests are running before that)

sanguinariojoe (approved):

PR #77, opened 2023-06-13
Improve decomposeString and fix other various bugs and warnings

AlexWKinley:
The main thing this commit does is get rid of a lot of the usage of char arrays and replace it with c++ strings. This means that decomposeString can handle inputs and outputs on length larger than 10, and also avoids a lot of raw string copies that are both dangerous and cause some compilers to produce warnings. A new test has been added that compares the output of the new decomposeString function with the old implementation to try and make sure they match.

A few errant bugs have been fixed, and a little work has been done to try and improve compile times. Mainly moving the step of solving a 6x6 system of equations into it's own single function in Misc.cpp prevents Body.cpp and Rod.cpp from each compiling the same code twice, and that code specifically takes a while to compile. I've also done some cleaning up of imports to try and help compile times. 
Additionally I've gone through and addressed some compiler warnings, MoorDyn produced quite a few warnings and decreasing the warning noise may help prevent useful compiler warnings from being missed.

This branch breaks the bodies_and_rods test because now `Body1Pinned` gets correctly parsed and the Rod is actually pinned to the body instead of being fixed, which along with however the physics are incorrect causes the test the fail. If desired I can comment out that test so that the CI passes until a more robust solution is available. I'm currently working on trying to find a solution to the bodies and rods problems, some of which may be ready to be integrated in the near future, but is generally a slow process.

### DIFF
--- a/source/Misc.hpp
+++ b/source/Misc.hpp
@@ -951,14 +951,13 @@ typedef struct _OutChanProps
 { // this is C version of MDOutParmType - a less literal alternative of the NWTC
   // OutParmType for MoorDyn (to avoid huge lists of possible output channel
   // permutations)
-	char Name[10];  // "name of output channel"
-	char Units[10]; // "units string"     - should match UnitsSize in
-	                // MoorDyn.cpp (should turn into def)
-	int QType;      // "type of quantity - 0=tension, 1=x, 2=y, 3=z..."
-	int OType;      // "type of object - 1=line, 2=connect"
-	int NodeID;     // "node number if OType=1.  0=anchor, -1=N=Fairlead"
-	int ObjID;      // "number of Connect or Line object", subtract 1 to get the
-	                // index in the LineList or ConnectList
+	string Name;  // "name of output channel"
+	string Units; // "units string"
+	int QType;    // "type of quantity - 0=tension, 1=x, 2=y, 3=z..."
+	int OType;    // "type of object - 1=line, 2=connect"
+	int NodeID;   // "node number if OType=1.  0=anchor, -1=N=Fairlead"
+	int ObjID;    // "number of Connect or Line object", subtract 1 to get the
+	              // index in the LineList or ConnectList
 } OutChanProps;
 
 // --------------------------- Output definitions
--- a/source/Misc.hpp
+++ b/source/Misc.hpp
@@ -569,7 +569,7 @@ typedef enum
  * Uses Eigen::ColPivHouseholderQR which means it doesn't
  * have any particular constraints on the matrix properties.
  * Has high accuracy and good speed.
- * 
+ *
  * @param mat 6x6 Matrix (M in M * a = b)
  * @param vec 6x1 Vector (b in M * a = b)
  * @return vec6 Resulting solution (a in M * a = b)
@@ -626,9 +626,11 @@ inline mat
 getH(vec r)
 {
 	mat H;
+	// clang-format off
 	H <<     0,  r[2],  r[1],
 	     -r[2],     0,  r[0],
 	     -r[1], -r[0],     0;
+	// clang-format on
 	return H;
 }
 
@@ -730,6 +732,7 @@ RotZ(real rads)
 	return R;
 }
 
+// clang-format off
 // Create the Euler rotations of the type RotXYZ, RotXZX, RotZYX...
 #define MAKE_EULER_ROT(a,b,c)                                                  \
 inline mat Rot ## a ## b ## c(real a1, real a2, real a3)                       \
@@ -740,6 +743,8 @@ inline mat Rot ## a ## b ## c(vec rads)                                        \
 {                                                                              \
 	return Rot ## a ## b ## c (rads[0], rads[1], rads[2]);                     \
 }
+
+// clang-format on
 
 MAKE_EULER_ROT(X, Y, X)
 MAKE_EULER_ROT(X, Y, Z)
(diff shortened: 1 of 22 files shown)


source/Misc.hpp
```diff
@@ -574,6 +563,20 @@ typedef enum
...
+ * Uses Eigen::ColPivHouseholderQR which means it doesn't
+ * have any particular constraints on the matrix properties.
+ * Has high accuracy and good speed.
+ * 
+ * @param mat 6x6 Matrix (M in M * a = b)
+ * @param vec 6x1 Vector (b in M * a = b)
+ * @return vec6 Resulting solution (a in M * a = b)
+ */
+vec6
+solveMat6(const mat6& mat, const vec6& vec);
```
sanguinariojoe, line 578: I suppose here we can create a template function called solveLinSys or something like, with 2 specializations for mat & vec and for mat6 & vec6. That way it will look cleaner, don't you think.

I am not stopping the merge for that anyway

source/Misc.hpp
```diff
@@ -826,7 +829,7 @@ typedef struct _FailProps
 const int nCoef = 30; // maximum number of entries to allow in nonlinear
                       // coefficient lookup tables
 
-typedef struct
+struct EnvCond
```
sanguinariojoe, line 832: In general the typedef is preferred, right? Basically because in C this way you have to refer to this type as `struct EnvCond` instead of `EnvCond` directly

source/Misc.hpp
```diff
@@ -975,20 +995,20 @@ typedef struct _OutChanProps
...
-const int VelY = 5;
-const int VelZ = 6;
-const int AccX = 7;
-const int AccY = 8;
-const int AccZ = 9;
-const int Ten = 10;
-const int FX = 11;
-const int FY = 12;
-const int FZ = 13;
+// const int Time = 0;
```
sanguinariojoe, line 998: Just remove the commented ints

source/Seafloor.hpp
```diff
@@ -1,13 +1,13 @@
 /**
-* Copyright (c) 2023, Kelson Marine Co.
-* 
-* This class is used to represent a three-dimensional
-* seafloor for use in a MoorDyn simulation.
-/
+ * Copyright (c) 2023, Kelson Marine Co.
```
sanguinariojoe, line 2: I did not notice. The copyright notice should incorporate the license:

```
/*
 * Copyright (c) 2023, Kelson Marine Co.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are met:
 *
 * 1. Redistributions of source code must retain the above copyright notice,
 * this list of conditions and the following disclaimer.
 *
 * 2. Redistributions in binary form must reproduce the above copyright notice,
 *    this list of conditions and the following disclaimer in the documentation
 *    and/or other materials provided with the distribution.
 *
 * 3. Neither the name of the copyright holder nor the names of its
 *    contributors may be used to endorse or promote products derived from
 *    this software without specific prior written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
 * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
 * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
 * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
 * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 * POSSIBILITY OF SUCH DAMAGE.
 */
```

It does not matter much, but in general is preferred to assign the copyright to the developers, not the companies. Up to you

sanguinariojoe:
Yeah, please, comment out the test until it is fixed

I also added some comments about low hanging fruit.

source/Misc.hpp
AlexWKinley: The `struct EnvCond` syntax for declaring a struct in c++ behaves nearly exactly the same as `class EvnCond` and you don't have to prefix the type with struct to use it. The other advantage of this syntax is that it allows forward declarations which happens in WaveGrid.hpp to avoid some unnecessary imports there which may help compilation in some cases.

source/Misc.hpp
AlexWKinley: I only didn't remove them because above and below there are some other commented out things about the out params that provide some potentially useful context and information. But I'll try and move that documentation parts to more useful locations and remove the commented out code.

source/Misc.hpp
sanguinariojoe: Your point seems perfectly reasonable

source/Misc.hpp
sanguinariojoe: We'll do this on the future

PR #88, opened 2023-06-26
Updates to rod dynamics

RyanDavies19:
Updates and improvements to rod dynamics. These changes reflect the rod dynamics in the OpenFAST version of MoorDyn, they have been verified against the MoorDynF code.


RyanDavies19:
Mistake, this was meant for the v2 branch

PR #89, opened 2023-06-26
Rod Dynamics Updates

RyanDavies19:
Updates and improvements to rod dynamics. These changes reflect the rod dynamics in the OpenFAST version of MoorDyn, they have been verified against the MoorDynF code.

I'm still getting the hang of how to do pull requests, so let me know if something is wrong here.

PR #90, opened 2023-06-27
Bodies and Rods Fixes

### LINKED ISSUE #70: Error introduced in bodies_and_rods test
In v2: Our team has observed that d0b09d27 causes `bodies_and_rods` to no longer converge during the IC stage. Furthermore, when MOORDYN_SSE is set to AVX2, the system actually diverges completely at `t=0.0003`s. This divergence is first caught in Line1, which connects Body1 and a fixed point. The two rods in the model appear to be linked to Body1 as well. This bug also persists in the latest commit ebcac61d.

We haven't determined the root cause of this, but it appears to be due to the changes to the mathematics in Misc.cpp, Misc.hpp, or Rod.cpp, as the input file was not changed in this commit. The issue persists even when the time-step is reduced to be quite low (0.000001), so it does not appear to be a minor stability thing.

AlexWKinley:
# Fixing Bodies and Rods

These changes should fix most (hopefully all) of the major issues with bodies and rods working properly. Closes #70 

The biggest change to MoorDyn is an overhaul of the kinematics for the 6DOF objects, bodies and rods. 
The full math of this change is detailed in the 6DOF kinematic section below. 
The overall highlights are that instead of euler angles, bodies and rods now use quaternions to represent their rotation. 
The integration of angular velocity in MoorDyn turned out to be incorrect (see 6DOF kinematics section), and this has been addressed by using the correct differential equation relating angular velocity and quaternion components. 
This PR also includes a number of new tests that hopefully ensure that bodies and rods are working correctly as well as prevent them from running into issues in the future.

## Smaller Changes/Fixes

A number of other fixes and changes have been made that contribute to the correctness of this code.

### Mass translation/rotation

The translate and rotate mass methods have been, as far as I can tell, brought to a fully working state. 
The general form stayed the same, but it turned out that even my correction to `getH`  was incorrect (I misread the fortran code). 
Additionally the translate mass methods have been made definitely correct.
To verify that these functions work properly there are tests in `tests/math_tests.cpp` that test that the functions produce the correct results in known cases.

### Rod Partial Submergence

Rod partial submergence now uses the same code that lines do. This fixes an issue where if the B end was lower than the A end, the buoyancy forces would still be applied on the A end. A fix for the same issue also change exists in #89, although that code more explicitly handles the inverted rod case.

### Rod Inertia

So far as I can tell, MoorDyn assumes that Rods have uniform density along their length. 
This means that the center of mass and moment of inertia of the rod itself (ignoring added mass from water), is analytically known.
Now rod inertia is calculated using the analytical equations for the moment of inertia of a rod at it's end. 
The added mass still adds inertia through translate mass.

### Rod Added Mass

Rod added mass has been fixed to be calculated correctly, volume of the end segments was unnecessarily halved. 
It seems like Ryan also fixed this issue in #89.
A test has been added to `test/rods.cpp` that check this by putting a buoyant rod with no drag and a Ca of 1 sideways 10m underwater, and watching it's position over the next ~2.3 seconds to ensure that it matches the analytical positions based on the buoyancy force and inertia from mass and added mass.

### Body VTK Output

Body VTK output has been fixed to correctly order rotation and translation, as well as to specify the correct rotation. 
This is mostly the fault of the VTK api being confusing.

### Medium Performance Improvements for Rods and Bodies

I added a benchmark that drops a few rods in the water. 
Profiling indicated that one of the major performance issues was the use of `Eigen::Seq`. 
Frequently throughout the code, `Eigen::Seq` is used in places where the indexes are known at compile time.
Without the use of `Eigen::fix<N>()`, the compiler does not seem to make use of this additional information to optimize the code.
Fortunately, Eigen provides a number of block operations on vectors and matrices that can take compile time indices. 
For vectors `my_vec.head<3>()` gets the first 3 elements of the vector, and `my_vec.tail<3>()` gets the last 3 elements of the vector.
For matrices, there are a number of block operations, including methods to get corners of the matrix.
`my_mat.topLeftCorner<3, 3>()` gets the 3x3 matrix from the top left corner, and there are other methods for the other corners.
[Eigen documentation on block operations](https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html).

Converting some of the highly used locations over to this api where possible amounted to I believe about a 30% performance improvement for rods, and probably also some gains for bodies. 

## Unit Testing

The other major change is the addition of a true unit testing framework, in this case Catch2. I decided to put the source for a recent stable version of Catch2 inside of the MoorDyn repo. This means that CI jobs shouldn't have to be changed, and avoids people having to download and install Catch2 themselves.

This should make it easier to test the actual code of MoorDyn more easily. I don't think most of the existing tests that mostly just run MoorDyn using various input files need to be changed over, they seem fine as they are. But I've changed testDecomposeString over to using Catch2, as well as added a new file `math_tests.cpp` that uses Catch2 to test some of the math functions inside if Misc such as translate mass. I think these sorts of unit tests are probably worthwhile to continue to add for complicated and/or critical functions to ensure that they work how we think they work, and that they continue to work in the future.

# 6DOF Kinematics

This is a long-winded and overly detailed description of the kinematic calculations done by bodies and rods in MoorDyn. 
The key points are that the way MoorDyn calculates acceleration from forces was mostly correct, but the way that MoorDyn calculates rotation from angular velocity was fundamentally incorrect, and that to correct it while avoiding gimbal lock, we have chosen to use quaternions. 

MoorDyn exclusively uses the world inertial coordinate frame for positions, forces, velocities, etc. 
The caveat to this is that positions are not at the center of mass, which creates the requirement for a 6x6 mass matrix.
From what I can tell, this is a fairly unique combination of requirements. 
In particular, ocean vessel dynamics are normally computed in a body fixed frame, where positions, velocity, forces, etc, are all relatively to the objects (generally a boat) coordinate frame.

Fortunately, MoorDyn's situation generally simplifies matters by avoiding needing deal with the fictitious forces that arise from a non-inertial reference frame.
The version of the newton-euler equation for this situation, where the center of mass is offset from the body positions by some vector $r$, but all vectors are in world coordinates is

```math
\begin{bmatrix}
F \\
\tau
\end{bmatrix}
=
\begin{bmatrix}
m & -m  H(r) \\
m H(r) & I_g + H(r)^T m H(r) 
\end{bmatrix}
\begin{bmatrix}
\dot{v} \\
\dot{\omega}
\end{bmatrix}
+ 
\begin{bmatrix}
0 \\
\omega \times ((I_g + H(r)^T m H(r))\omega)
\end{bmatrix}
```

$F = \text{net force}$

$\tau = \text{net moments}$

$m = \text{3x3 mass matrix at CG}$

$v = \text{velocity}$

$\omega = \text{angular velocity}$

$H(r) = \text{getH(r), the cross product operator, should produce a matrix where } H(r) a = r \times a$

That second term is the gyroscopic moment, which currently is entirely neglected, making the assumption that nothing is rotating fast enough to warrant its inclusion.

This set of conventions I believe differs slightly from Sadeghi and Incecik. They define $r$ to be from the point mass to the body position point. 

This is of course only one of two required differential equations to define equations of motion. The other one is the one that relates position and rotation to velocity and angular velocity.

For velocity the relation is simple due to using world coordinates for everything

$$\dot{p} = v$$

For rotation, things are more complicated. What MoorDyn previously assumed was that our rotation using euler angles $\phi, \theta, \psi$ was related to our angular velocity by 

$$
\begin{bmatrix}
\dot{\phi} \\
\dot{\theta} \\
\dot{\psi}
\end{bmatrix}
= \omega
$$

This is an incorrect assumption due to the fact that euler angles are a sequential series of rotations.
Consider an object that rotates at 90 deg/second around the y axis for 1 second. 
Then stops and rotates at 90deg/second around the x axis for one second, and then the same for the z axis.
The object is now rotated 90 degrees around the x axis relative to its original orientation. 

The above equation would give us that $\phi = 90$, $\theta = 90$, and $\psi = 90$. The rotation those euler angles represent is not a 90 degrees rotation around the x axis (using a XYZ convention it represents a rotation of 90 degrees about the y axis and then 180 degrees about the z axis, using a ZYX convention it represents a 90 degree rotation around the y axis).

The correct equation relating rotation in euler angles and angular velocity (from [1]) is:

```math
\begin{bmatrix}
\dot{\phi} \\
\dot{\theta} \\
\dot{\psi}
\end{bmatrix}
= 
\begin{bmatrix}
1 & \sin(\phi)\tan(\theta) & \cos(\phi) \tan(\theta) \\
0 & \cos(\phi) & -\sin(\phi)  \\
0 & \sin(\phi) / \cos(\theta) & \cos(\phi) / \cos(\theta)
\end{bmatrix}
\omega
```

This equation will give correct results for many cases, but when $\theta$ is at or near $\pm90$ a discontinuity is approached and divide by zeros can occur. This is related to the phenomena known as gimbal lock [2]. There are a few potential solutions to this problem, including ignoring it if you have reason to believe that a pitch of $\pm90$ is not going to occur (such as if you are simulating a boat or aircraft under normal conditions). It is also possible to switch between two sets of euler angles with different discontinuity points. But a common and robust solution is to use quaternions, or more specifically unit quaternions, also sometimes called euler parameters. 

I won't go into all the math of quaternions, but essentially they are 4 dimensional complex numbers that nicely map onto rotations in 3D space. Eigen has built in support for quaternions meaning that most of the complicated math is hidden.

The thing worth noting is that with quaternions we need a different differential equation relating change in quaternion components to angular velocity. For an angular velocity given in a world coordinate system (as opposed to a body fixed coordinate system) we have [3]:

```math
\begin{bmatrix}
\dot{q_w} \\
\dot{q_x} \\
\dot{q_y} \\
\dot{q_z}
\end{bmatrix}
 = \frac{1}{2} 

\begin{bmatrix}
0 \\
\omega_x \\
\omega_y \\
\omega_z
\end{bmatrix}
  \otimes q
```

Where $\otimes$ represents the Hamilton product of quaternions [4], which is conceptually equivalent to multiplication of complex numbers, but with 3 different imaginary constants.

## References
[1] Fossen, Thor I. Handbook of marine craft hydrodynamics and motion control. page 25 John Wiley & Sons, 2011.

[2] https://en.wikipedia.org/wiki/Gimbal_lock

[3] https://www.ashwinnarayan.com/post/how-to-integrate-quaternions/

[4] https://en.wikipedia.org/wiki/Quaternion#Hamilton_product


sanguinariojoe:
First, I shall definitely move all the seq calls to .head and .tail then!
Second, I also wanted to move to quaternions. I really like it much better. However, that is making this code even more different from the Fortran one.

I would merge this right away and start working from here (e.g. https://github.com/FloatingArrayDesign/MoorDyn/pull/91 is still a problem here).

Let's wait for the feedback of @mattEhall and Ryan though, because this is clearly a one way path.

sanguinariojoe:
Hey @AlexWKinley , I have added several fixes to this. You have a [PR](https://github.com/KelsonMarine/MoorDyn_Public/pull/1) on your repo

sanguinariojoe:
BTW, the wheels building is failing:

https://github.com/sanguinariojoe/MoorDyn/actions/runs/5463096677/jobs/9943358544

mattEhall:
Good catch on the bad assumptions in the original formulation for rotations. Fully agree. It was just an approximation that worked decently for the small rotation scenarios that were most common in past work. This upgrade to use quaternions sounds great.

There were a couple details I wasn't sure of, we have some better free surface hydrodynamics handling to include from the Fortran code, and Ryan tells me that some tests are giving odd results -- but I'm onboard with making this switch to quaternions and then dealing with these additional details afterward.

RyanDavies19:
Going off what Matt said, I am working on a couple fixes that should resolve the issues, they mostly pertain to free rods. Once this PR gets implemented I will open a new one with those changes.

sanguinariojoe:
OK, I am merging then and cherry picking the changes [here](https://github.com/KelsonMarine/MoorDyn_Public/pull/2)

PR #91, opened 2023-06-28
Minimal test of lines attached to a rigid body

sanguinariojoe:
We have been detecting some issues on the OpenFOAM wrapper, so I created this test to grant it is not a problem on MoorDyn


sanguinariojoe:
Effectively, when the body has a non null COG errors happens. That might be related with the logic, so I must check it

PR #93, opened 2023-07-10
Rod dynamics and outputs updates

RyanDavies19:
These changes fix a few bugs that existed with rod dynamics, in particular for free rods. The biggest issue was the rod Mass matrix was double counting the node masses. Other issues involved horizontal rods rotating when sinking/falling. They have been verified against MoorDynF and analytical solutions. 

The other set of changes is output flag standardization with MoorDynF. This is in conjunction with an update to MoorDynF that will come soon, and it makes it so both versions accept the same output flags and produce the same output results. Much of it was to bring this version up to date with what the documentation says it can do, and then there are some added features. These changes have also been verified with the changes in MoorDynF for all of the output channels.

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -307,29 +307,62 @@ Body::setDependentStates()
 real
 Body::GetBodyOutput(OutChanProps outChan)
 {
+
+	vec3 rotations = Quat2Euler(r7.quat);
+
 	if (outChan.QType == PosX)
 		return r7.pos.x();
 	else if (outChan.QType == PosY)
 		return r7.pos.y();
 	else if (outChan.QType == PosZ)
 		return r7.pos.z();
+	else if (outChan.QType == RX)
+		return rotations[0];
+	else if (outChan.QType == RY)
+		return rotations[1];
+	else if (outChan.QType == RZ)
+		return rotations[2];
 	else if (outChan.QType == VelX)
-		return v6[0];
-	else if (outChan.QType == VelY)
-		return v6[1];
-	else if (outChan.QType == VelZ)
-		return v6[2];
-	// else if (outChan.QType == Ten )  return  sqrt(Fnet[0]*Fnet[0] +
-	// Fnet[1]*Fnet[1] + Fnet[2]*Fnet[2]);
-	else if (outChan.QType == FX)
-		return F6net[0]; // added Oct 20
-	else if (outChan.QType == FY)
-		return F6net[1];
-	else if (outChan.QType == FZ)
-		return F6net[2];
-	else {
-		LOGWRN << "Unrecognized output channel " << outChan.QType << endl;
-		return 0.0;
+ 		return v6[0];
+ 	else if (outChan.QType == VelY)
+ 		return v6[1];
+ 	else if (outChan.QType == VelZ)
+ 		return v6[2];
+ 	else if (outChan.QType == RVelX)
+ 		return v6[3]*180.0/pi;
+ 	else if (outChan.QType == RVelY)
+ 		return v6[4]*180.0/pi;
+ 	else if (outChan.QType == RVelZ)
+ 		return v6[5]*180.0/pi;
+ 	else if (outChan.QType == AccX)
+ 		return a6[0];
+ 	else if (outChan.QType == AccY)
+ 		return a6[1];
+ 	else if (outChan.QType == AccZ)
+ 		return a6[2];
+ 	else if (outChan.QType == RAccX)
+ 		return a6[3]*180.0/pi;
+ 	else if (outChan.QType == RAccY)
+ 		return a6[4]*180.0/pi;
+ 	else if (outChan.QType == RAccZ)
+ 		return a6[5]*180.0/pi;
+ 	else if (outChan.QType == Ten)  
+ 		return  sqrt(F6net[0]*F6net[0] + F6net[1]*F6net[1] + F6net[2]*F6net[2]);
+ 	else if (outChan.QType == FX)
+ 		return F6net[0];
+ 	else if (outChan.QType == FY)
+ 		return F6net[1];
+ 	else if (outChan.QType == FZ)
+ 		return F6net[2];
+ 	else if (outChan.QType == MX)
+ 		return F6net[3];
+ 	else if (outChan.QType == MY)
+ 		return F6net[4];
+ 	else if (outChan.QType == MZ)
+ 		return F6net[5];
+ 	else {
+ 		LOGWRN << "Unrecognized output channel " << outChan.QType << endl;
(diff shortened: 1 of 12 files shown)


RyanDavies19:
Also just an FYI to all, I am working on changing all mentions of 'connections' to 'points' per Matts instructions. Going forward points are what should be used. There should be a PR with this next week.

sanguinariojoe:
Looks OK! I am merging and fixing the testing pipelines...

This time for good! I promise!

AlexWKinley:
I've done some work with the Rod dynamics changes in this PR, and I have some questions and notes. Firstly I understand why you thought I double counted the node masses, but I had removed the node masses from the `M` matrices, so the actual mass of the rod was only counted once. The reason I had made that change is because the current method of computing the moment of inertia is not particularly accurate, and using the known equations for the moment of inertia of a cylinder from its end will always give accurate results. At some point during my testing I had verified the rod moment of inertia by comparing against the equations for the period of a [physical pendulum](http://hyperphysics.phy-astr.gsu.edu/hbase/penrod.html#c1). 

The current way that MoorDyn computes the moment of inertia (in particular the moment of inertia around the axes orthogonal to the axis through the rod) is mainly through using a lumped mass approximation, as well as having an additional I_r term (the non-parallel axis term as it is called in the code). For the case of a rod with no added mass, this does not produce the correct results.
MomentOfInertia

This graph shows the calculated moment of inertia for a rod 2 meters long, with a diameter of 0.3m and a mass of 20kg. The exact value can be verified [using this calculator](http://hyperphysics.phy-astr.gsu.edu/hbase/icyl.html)  (the value being calculated is $I_{\textbf{end diameter}}$). As well as graphing the exact value and the value MoorDyn calculates, I've also included the value MoorDyn calculates without the additional I_r term. Interestingly, that value seems to actually do a better job of approximating the exact solution. Because of the fact that MoorDyn overestimates the moment of inertia for rods, and the exact value is easily computed, I had chosen to forgo the lumped mass approximation of moment of inertia entirely and just use the exact value. This meant that the lumped mass approximation was left to just account for the hydrodynamic added mass. 

Another aspect of this PR that I have questions about is the buoyancy calculations. Acknowledging that computing buoyancy for a partially submerged cylinder is no easy task, I don't understand why MoorDyn computes buoyancy forces the way that it does. Where the end faces exert buoyancy forces along the axis of the rod, and the nodes experience buoyancy forces orthogonal to the rod axis. For this reason, I was impressed when my testing found that this formulation actually produces reasonable results. My testing involved having a rod calculate the forces on it when partially submerged at a variety of angles. The setup looked basically like this, for each case the center of the rod is in the same location, slightly below the water surface, but the angle of the rod from vertical is varied from 0 to $\pi$
image

I looked at vertical buoyancy force as well as the torque on the rod (around the y axis because the rod was being rotated towards the +x axis).
ZFb
ytb

The ground truth is based on some code I have written that exactly solves for the wetted volume and center of buoyancy of a cylinder intersecting the water plane. I based those calculations on the equations given in [Apman, R. P. (1961)](https://core.ac.uk/download/pdf/228666464.pdf). 
MoorDyn seems to do a good job approximating the truth, although with a low resolution it certainly is not perfect.

Where MoorDyn did a much worse job in my testing, in accurately calculating the linear and angular accelerations. This is a result of the incorrect calculation of the moment of inertia. I compared the accelerations in the z and x axis, and the angular acceleration around the y axis, with the ground truth values for the current implementation of mass, as well as the same with my corrected mass and moment of inertia calculations.
az
ax
aay

At least in these tests, having a correctly calculated moment of inertia gave results that were reasonably close to the ground truth. 

That all being said, it's likely that my code for analytical buoyancy forces will end up in a PR to this repo in the near future. Some rough benchmarks seem to indicate that it is not noticeably slower than the existing buoyancy calculations. I still have the open question of if there is a good way to use the calculations that code does to get accurate segment submergence (or perhaps enable even more accurate ways of modeling forces on partially submerged rods?). Additionally, handling waves beyond a horizontal plane approximation is something I haven't thought too much about, any thoughts or ideas are welcome. 

I'd love to get some more insight on the way that MoorDyn currently handles buoyancy forces. I haven't come across any formulation like that before and would be very curious to read more about it.

mattEhall:
I've now responded about the buoyancy calculation here https://github.com/FloatingArrayDesign/MoorDyn/pull/97#issuecomment-1641249047

mattEhall:
Hey sorry I'm late on this, but is there a way we could keep the structural mass calculated per-node just like the added mass is, but correct the errors that were causing it to be overestimated?

The value of calculating per node is that it allows the inertial component of internal structural forces in the rod to be computed. And I imagine the computational burden is negligible because we already have to sum up node masses because of added mass.

AlexWKinley:
It's probably not out of the question to do some sort of correction step, you could likely just compute the difference between the lumped mass and exact solution and use that. I am interested in what you mean by "inertial component of internal structural forces". I don't follow what you mean by that, and looking at the MoorDynF code I don't think I see anything that seems like what you are referring to. I see some mention of inertial loads, but those use the full 6x6 `M6net` matrix. Without being fully sure what those forces are, or how they get calculated, I will note a few things.

1. The change to using the exact cylinder moment of inertia formula only changes the `Rod%M` matrices, which are the 3x3 mass matrices for each of the nodes on the rod. (With my change, the M matrices just contain the added mass tensor for each node, and not also the physical mass)
2. At any point it is straightforward to compute the lumped mass of the nodes and reconstruct what the node mass matrices would be. This obviously isn't computationally efficient if it is being done very frequently. But if it's something that is calculated every time step for every node then it could happen inside of the node loop in `Rod_DoRHS` where you already have computed the value `m_i` to calculate the node weight. 

If none of that would resolve the issue I can definitely look into the idea of a calculating a moment of inertia correction to account for the difference between the lumped mass and exact moments of inertia.

RyanDavies19:
It's worth noting that I think the radial moment of inertia calc at the for rods in MoorDynF is a bug that needs to be fixed. The formula is based on the Hydrodyn document @mattEhall shared in #97 (page 4 non tapered members). MoorDynF takes each segments mass as the whole rod mass, rather than a fraction of the mass as indicated in the theory paper.

mattEhall:
Hey Alex, that inertial structural force would be for example the tension force that would develop in a cylinder if you're swinging it in an arc about one of its ends. Or in general the internal force that would arise when any external force is causing the rod to accelerate. Calculation of that force isn't currently implemented. It's a future possibility I've had in mind that I chose the formulation to support.

I hear you about your point 2 to reconstruct local mass matrices. I just think it's simpler to use local mass matrices from the start since we already have them for added mass and it aligns more closely with how Line objects work.

PR #96, opened 2023-07-18
WIP: Improvements for the implicit time schemes

sanguinariojoe:
I have been revisiting the implicit methods to make them more attractive. They are totally able to get the best accuracy, but in terms of speed they are still not able to outperform RK2. They are close now though...


sanguinariojoe:
I started including the CFL number to try to keep stability on the implicit schemes

PR #97, opened 2023-07-18
Improve Rod hydrodynamics and moment of inertia calculation

AlexWKinley:
This PR was formerly a PR for exact submergence calculations, that code can be found at https://github.com/KelsonMarine/MoorDyn_Public/tree/exact_submergence if people are interested.

Now this PR improves how rods handle partial submergence, by taking into account the water height at each node in more cases. 
This PR also includes a correction term for the moment of inertia that exactly compensates for the lumped mass approximation of moment of inertia. I also fixed the fact that changing the meaning of phi to be relative to the z axis instead of relative to the x axis requires the additional moment of inertia matrix to have a different ordering of terms. 

<details>
  <summary>Old PR Message</summary>
This PR includes a closed form, exact in the planar water surface case, formulation for rod buoyancy forces and the buoyancy moment. The calculations are independent of the number of segments that the rod has, and instead computes the wetted volume, and centroid of that volume for the entire cylinder of the rod. See my comments in PR #93  for more details on the motivation for this work. Also mentioned there, and implemented here, is that I have gone back to using my exact mass and moment of inertia calculation for the entire rod. This allows the computed accelerations due to buoyancy to match the expected values. 

## Submerged Volume Formulation

The equations for the wetted volume and center of buoyancy are based on the equations given in [Apman, R. P. (1961)](https://core.ac.uk/download/pdf/228666464.pdf). In particular, the equations for case II on page 10, along with the derivation and definition of terms given in the appendices. That equation allows the calculation of the wetted volume and center of buoyancy for any cylinder whose bases intersect the water plane (or the edge of the base intersects the water plane). This makes it possible to turn any case of a partially submerged cylinder into some combination of either or of a fully submerged cylinder and a partially submerged wedge whose bases intersect the water plane. These cases are illustrated below (in the order that they appear in `AnalyticalBuoyancyCalculator::calculateBuoyancy`).
 
submergence_cases

The angle $\varphi$ is an angle between 0 and $\pi$ that describe the height of the water plane on the base of a cylinder.
submergence_phi

My C++ implementation of this algorithm is based on a python implementation that I previously wrote and semi-rigorously tested in a number of ways. That code has been used to generate the relatively large number of test cases for the c++ code. As a simple gut check for those test cases, this is a (admittedly cluttered) visualization of some of those test cases. Where the cylinders all have one base at the same location, and the direction of the cylinders is varied. Here the darker circle is the center of the circle and the lighter x is the computed center of buoyancy of the cylinder.
rod_cob_bottom_y_0 2

I have retained the existing VOF calculations for use by the other hydrodynamic forces, although `zA` is no longer computed since it was only used for buoyancy calculations. So far as I can tell, this new code has a neutral to positive effect on performance. I've done my best to avoid unnecessary computations, and especially trig function evaluations.  And it has the advantage of only needing to run once for the whole cylinder, instead of per-segment/node. 

I'm happy to do my best to answer questions about the way that submerged volume and center of buoyancy are calculated. I haven't given much thought to how this could be further incorporated with the calculations for the other hydrodynamic forces, although as seen in my comments on in #93 the existing calculations are reasonably accurate so maybe it's not necessary. 

What likely worth more thought and deliberation is handling variation in surface height. The existing version of the rod buoyancy calculations do take the local variation in surface height into account (at least to some extent, I have not done much testing of that). Currently, my code just uses the water surface height at the top end of the rod. It should be possible to convert the current code to be per-segment if desired. Then it could compute the wetted volume and center of buoyancy for each segment of the rod. The performance implications of that maybe a slightly more dramatic for rods with many segments that are partially submerged. I'm interested in peoples thoughts on the matter, since the top point approximation may be sufficient for something like an upright buoy, but would not do well for something like a long buoyant cylinder lying on the waters surface. 

</details>


mattEhall:
Hi Alex, don't hesitate to email me or tag me if you need my attention. There are two reasons for the distributed calculation of hydrostatic forces on rods: (1) compatibility with waves and (2) compatibility with potential future features related to flexibility.

The first reason is probably enough to worry about for now. The distributed hydrostatic approach, while dependent on the discretization (like most things in MoorDyn), worked quite well in cases with waves. At least it does in MoorDyn F; it may have not gotten there in MoorDyn C yet. I don't think an approach that computes buoyancy based on displaced volume can work well when waves are involved because there is a fundamental clash with the calculation of the dynamic pressure term.

sanguinariojoe:
I agree with you two. I think we need something discretized, like @mattEhall already discussed. However, the exact approach of @AlexWKinley is for sure beating any approximation we are considering (either here or in MoorDynF).

I think @AlexWKinley just need to call his buoyancyCalculator segment by segment. Would you do that?

mattEhall:
> I think @AlexWKinley just need to call his buoyancyCalculator segment by segment. Would you do that?

That wouldn't work. Alex's method is based on displaced volume rather than surface pressure; they're fundamentally different. 

Does anyone have a problem with the discretized approximate approach for rod buoyancy?   (Keeping in mind that all other aspects of rod hydrodynamics at the free surface are equally approximate...)

sanguinariojoe:
AFAIK these kinds of problems are usually tackled splitting the problem in dynamic pressure and hydrostatic one. Buoyancy is indeed the force resulting from the application from the hydrostatic force.

Obviously, volume displaced and surface hydrostatic pressure gives the same result. I write the demo because I think @AlexWKinley asked about it.

Starting from the pressure force

image
image

So we can include the interface caps between the segments, because their integrals are balanced:

image

On the other hand, for an hydrostatic field we can apply the Gauss theorem (for simplicity I assume that x = (0,0,0) is on the free surface):

image
image
image

--------------------

Thus I strongly recommend to address the buoyancy by the @AlexWKinley function, adding the dynamic pressure forces afterwards

mattEhall:
@sanguinariojoe I'm having trouble seeing where the dynamic pressure term is handled in the equations you wrote. Also, making the assumption that 0,0,0 is on the free surface seems like a major problem if we're talking about waves (which are defined by the free surface moving). 

@AlexWKinley the issues arise when considering waves. Waves create a dynamic pressure field in the water, which acts on the cylinder ends in a non-equal way. (Dynamic pressure on the sides is handled by the 1+Ca in the Froude Krylov / inertial transverse force term, and we assume slender cylinders so it's less of an issue.)  So surface pressure and displaced volume approaches will give different answers when waves are involved. One of those answers is closer to the truth than the other ;)

If people have too coarse a discretization to get accurate hydrostatics, I'm pretty sure it will also be too coarse to get accurate hydrodynamics. That's why I see this as a non issue.

There is a large body of practice behind the approach I implemented (from OpenFAST) and it has also compared quite well with experimental and OrcaFlex results. I don't think you'll be able to get that with the other approach (but feel free to prove me wrong of course).

In terms of new features, I don't have plans anytime soon for flooded or tapered members, but that could change in the future. One additional benefit of the surface pressure approach is that -- because it matches how water forces actually work -- it allows true calculation of the internal structural forces in a member (for example the axial compressive force).

sanguinariojoe:
That's because I am considering just the hydrostatic one :-p
The dynamic pressure shall be added the same way it was done before. 0,0,0 at the free surface is just a way to simplify the notation, because they do not contribute to the gradient.

AlexWKinley:
@mattEhall I'll defer to your expertise and experience with regards to method of computing hydrostatic and hydrodynamic forces. It's worth noting that currently in this version of MoorDyn, as well as in any version of MoorDynF that I can see in the openfast repo, rods seem to have only partial support for varying wave height and things like non-monotonically varying submergence, such as hogging/sagging behavior (in this repo the water elevation at the top end of the rod is used for calculating the surface pressure on the sides of the rode, and in MoorDynF the way that h0 is calculated and used assumes that the central axis of the rod only passes through the water surface a maximum of 1 times) . In both cases these limitations seem to be due to the implementation and not fundamental limitations of the formulation. My short term plan is to switch back to the existing buoyancy calculations and see if I can address those implementation limitations. These exact wetted volume calculations will probably live in some other branch in the Kelson Marine MoorDyn_Public repo, if others would like to look through or experiment with that code. 

I'll be interested in the results of our discussion in #93 about moment of inertia calculations. I'll leave this PR open for the time being, and it can turn into the PR that includes any of my improvements to the surface pressure based buoyancy calculations, as well as whatever method we land on for making sure the moment of inertia is calculated accurately. 

I'll do my best to keep this thread posted with any results I have. Let me know if you see any issues with my current plan.

AlexWKinley:
@mattEhall This updated PR includes improved handling for local surface height changes. I do have two questions about the calculations to make sure I understand the math that is happening. 

When calculating the "sideways submergence" for each segment, the surface height at the node is used. During those calculations the `zA` term is calculated, which I believe is basically the depth of the center of buoyancy. When the water surface is not at z=0, it seems like this term should be adjusted so that `zA < 0` means the segment is under water and `zA > 0` means the segment is above water. The most obvious correction would be to replace `r[i][2]` with `r[i][2] - zeta[i]` when it's used to calculate `zA` (which happens in both the fully submerged and partially submerged cases). This ran into what seemed like an issue, where the net buoyancy force could exceed the maximum possible hydrostatic buoyancy force in cases where the top of the rod was below a lower part of the water surface and the lower end of the rod was below a higher part of the water surface. This creates a situation where `zA_top - zA_bottom` is larger than the actual vertical distance between the two ends of the rod. I wasn't sure what the best way to handle that was. On the one hand that may be an accurate model of what is occurring, but it seemed possible to me that this would double count dynamic pressure forces. My remedy for this problem was to instead replace `r[i][2]` with `r[i][2] - zeta_i` where `zeta_i` is actually the water surface height at the top rod node. This solution also probably runs into issues. I don't know if there's a know better way of handling this that should be implemented in MoorDyn. The Hydrodyn document suggests a procedure where each segment is calculated in a frame where the local water surface is a plane a z=0, and the nodes are translated and rotated as needed, but I don't know how applicable that is to the formulation that MoorDyn is using. 

And another just clarifying question, you mentioned that the dynamic pressure on the sides of a rod is handled by the (1+ Ca) term in the added mass force. Is that not true for the rod ends, which have both an added mass force using 1 + CaEnd, as well as a dynamic pressure? It's been a while since I've sat down and gone through the derivation of the way that MoorDyn calculates added mass, so I just figured I'd bring that up in case it wasn't intentional. 

Let me know if there are other things that I should address in this PR, or if there are any issues with the current code that need fixing.

RyanDavies19:
@AlexWKinley Hopping in here to say Matt is out of the office until 8/7 so he might not be able to respond to this right away. You are right about the end effect force being strange. It should be just CaEnd rather than 1 + CaEnd. This is a bug in the fortran code as well that I am making a PR for. I should have one for this code up soon with fixes to that as well as some other fixes that ensure this code still matches the Fortran code.

source/Rod.cpp
```diff
@@ -1263,37 +1271,47 @@ Rod::doRHS()
...
+		 * I_exact = 1/4 * mass * r^2 + 1/3 * mass * length^2
+		 * The lumped mass approximation can be found by using the fact
+		 * that a point mass distance r from the coordinate center had a moment
+		 * of inertia of m * r^2.
+		 * A rod of length L with N segments has N+1 nodes whose distance from
+		 * the rod end is given by dist(n) = (n-1) * (L/N). (Assuming that node
+		 * 1 is the first node)
+		 * If the rod has a total mass M, then the mass of the inner nodes is
+		 * M/N and the mass of the two outer nodes is M/2N.
+		 * We can ignore node 1 since it is at distance zero and so adds no
```
RyanDavies19, line 1297: this should say "we can ignore node 0" to match the indexing used

source/Rod.cpp
```diff
@@ -1263,37 +1271,47 @@ Rod::doRHS()
...
+		 * moment of inertia.
+		 * Therefore we can write the total moment of inertia as
+		 * (sum from n=2 to N of (M/N)*dist(n)^2) + M/2N * dist(N+1)
```
RyanDavies19, line 1300: similar to above, should be "(sum from n=1 to N of ..."

mattEhall:
@AlexWKinley, one more note on something I was chatting about with @RyanDavies19.

A few weeks ago you wrote "the way that h0 is calculated and used assumes that the central axis of the rod only passes through the water surface a maximum of 1 times", and it seems that you went ahead and changed that.

The current approach works well and the new approach is less consistent (and Ryan found a bug with it). I know you thought it was a limitation, but this h0 calculation is only significant for rods that are far from horizontal (in which case they'll only ever cross the water line once). So there's no problem than needs fixing as far as I can tell.

AlexWKinley:
In general, I agree that the dynamics of Rods seem reasonable as they are. I do think the way that `zA` is calculated current does lead to some significant issues with the presence of waves. It may be that the correct way to handle surface elevation variations with waves is some more general wave stretching technique. But with the general method that MoorDyn uses now, a vertical rod partially submerged in a wave peak or trough will experience unexpected buoyancy forces.

If a rod is vertical and partially submerged in a wave trough, only the bottom segment will experience a non-zero buoyancy force.  This force will be along the vector from the bottom of the rod to the top of the rod (the +z axis vector), with a magnitude given by:
`Ftemp = -VOF[i] * Area * env->rho_w * env->g * zA;`
Assuming that the bottom segment of the rod is entirely submerged, this buoyancy force is equivalent to the weight of water displaced by a cylinder of length `zA`. In other words the buoyancy force is what it would be if there was no wave. But since the rod is in a wave trough, it may be much less submerged than the distance between z=0 and `zA` implies. 

Similarly, a vertical rod partially submerged in a wave peak can experience a buoyancy force pulling it downwards when `zA` becomes positive but the bottom segment of the rod is still submerged. 

Like I said, it's possible the solution to this needs to be more general than just offsetting `zA` such that it is distance below the water.

@RyanDavies19 I'm on board with that plan. I'm interested to see what the change back to the h0 approach looks like, so I'll keep an eye out.

Once the moment of inertia change gets merged into MoorDyn through Ryan's PR, I'll close this PR.

mattEhall:
@AlexWKinley I hear what you're saying about the buoyancy force, but I think you'll find that the issue you described is taken care of by the presence of the dynamic pressure term. After considering this--or running simulations to test it if need be--let us know if you still think there's anything wrong. 

I may have been misleading earlier about wave stretching for dynamic pressure. It's been a while since I looked at it. Wave stretching is probably just relevant for wave kinematics. I think the dynamic pressure is already first-order correct as-is.

PR #101, opened 2023-07-27
Body and rod bug fixes

RyanDavies19:
Resolves a body added mass bug, a rod Aq bug, rod inertia bug, and a translateMass bug:

The following three bugs have been fixed in the fortran code and are being implemented in a PR soon.

1. The body added mass calculation was adding a volume to the mass, which is dimensionally incorrect, and was missing the water density and the Ca coefficient. To resolve this, the env variable needed to be loaded in the setup function rather than a separate function. To ensure continuity across the objects, I made the setEnv function setWaves (an inline function) and the env variable is now set in the setup function for each object.
2. The rod Aq bug was identified by others at NREL who work on the fortran side. The force labeled Froude-Krylov force is actually the long-wave diffraction force. The force labeled dynamic pressure is actually the Froude Krylov force. The 1 that was added to CaEnd was a mistake because it represents the FK contribution which had already been accounted for in the dynamic pressure calculation (now FK force). See Eq. (32) on page 51 of the HydroDyn manual for the detailed equations: https://www.nrel.gov/wind/nwtc/assets/downloads/HydroDyn/HydroDyn_Manual.pdf 
3. The rod inertia bug has been discussed in #93. The moments of inertia for each rod segment were calculated and summed, however each segment was given the whole rod mass. This has been resolved by dividing the mass by N. See page 4 (non tapered members) for theory: https://openfast.readthedocs.io/en/main/_downloads/39aafbb61663ae16bd1a7e80f305c20b/HydroDyn_Plan_TCF_Morison.docx

The translate mass bug seemed to be introduced in #90. The matrix in the getH function was the transpose of the one in the fortran code. I adjusted the ordering of things in the translate mass function accordingly and v2 cpp now matches the rod dynamics of v2 fortran.

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -59,8 +59,10 @@ Body::setup(int number_in,
             vec I_in,
             vec6 CdA_in,
             vec6 Ca_in,
+			EnvCondRef env_in, 
             shared_ptr<ofstream> outfile_pointer)
 {
+	env = env_in;     // set pointer to environment settings object
 	number = number_in;
 	type = type_in;
 	outfile = outfile_pointer.get(); // make outfile point to the right place
@@ -99,7 +101,7 @@ Body::setup(int number_in,
 	M0 = translateMass6(body_rCG, Mtemp);
 
 	// add added mass in each direction about ref point (so only diagonals)
-	M0 += mat6::Identity() * bodyV;
+	M0 += bodyV * bodyCa.asDiagonal() * env->rho_w; // Values are only non-zero if free body
 
 	// --------------- if this is an independent body (not coupled) ----------
 	// set initial position and orientation of body from input file
@@ -226,13 +228,6 @@ Body::initialize()
 
 	return std::make_pair(r7, v6);
 };
-
-void
-Body::setEnv(EnvCondRef env_in, moordyn::WavesRef waves_in)
-{
-	env = env_in;     // set pointer to environment settings object
-	waves = waves_in; // set pointer to Waves  object
-}
 
 void
 Body::setDependentStates()
--- a/source/Body.hpp
+++ b/source/Body.hpp
@@ -200,6 +200,7 @@ class Body final : public io::IO
 	 * @param I Inertia (diagonal matrix components)
 	 * @param CdA Drag coefficient
 	 * @param Ca Added mass coefficient
+	 * @param env_in Global struct that holds environmental settings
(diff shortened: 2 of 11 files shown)


AlexWKinley:
I have two small notes on these changes. I chose to go with the version of getH I did because it matched the form that you more frequently see in reference texts. My version is equivalent to the cross product operator https://en.wikipedia.org/wiki/Cross_product#Conversion_to_matrix_multiplication, which frequently is used when doing these sorts of calculations https://en.wikipedia.org/wiki/Newton%E2%80%93Euler_equations#Any_reference_frame. I'm happy with making both version of MoorDyn match, but I did have a test in tests/math_tests.cpp that checks for getH producing the cross product operator. This should be easily fixed by changing the test to check for cross product but with flipped sign. (I'm assuming that the other tests that check for the exact behavior of translateMass and translateMass6 pass, if they do not I can look into it further) @sanguinariojoe.

Additionally for the moment of inertia calculations I believe that equation for the moment of inertia of a rod segment effectively just adds the moment of inertia of the first segment N times, Given that the lumped mass calculations already compute an approximate moment of inertia for the entire rod, I'm not sure why that code is doing that. If it were trying to compute the total moment of inertia of the rod, it would have to account for the parallel axis theorem. PR #97 in its current state includes the correct calculation for ensuring the final moment of inertia due to the rod mass is exactly correct (along with a fairly detailed explanation of the derivation). Essentially you have to add the radial portion of the moment of inertia for the entire rod, and then subtract a correction for the fact that the lumped mass model overestimates the non-radial moment of inertia. @RyanDavies19

sanguinariojoe:
> This should be easily fixed by changing the test to check for cross product but with flipped sign.

I forgot to upload the changes... Done here: 0b784acfb71a71da64dc68daac0fb3e6066a5941

sanguinariojoe:
BTW, I would inform you guys that I am beginning my parental leave, so i will be way less responsive for a couple of months. I will try to keep it up anyway

RyanDavies19:
@AlexWKinley I think that sticking with the getH function as is now is a better option because it keeps the underlying physics of the Fortran and Cpp codes the same. The cross product method was not producing the same 6x6 mass matrix as the fortran code and was causing differences in body behavior between the models. 

Your moment of inertia calculations from #97 seem to be the best approach to me. I missed that it was just N times the first segment moment of inertia. I will discuss with others because implementing them would mean changing the fortran code as well.

Issue #108, opened 2023-08-16
Total Force and Moment Calculation Issue


soydanahmet:
Hi,

I'm trying to couple Mordyn v1.01.02 with my code and my input file is attached. My platform center of mass is 3.5, 0.5, 0.4744.

So X=3.5,0,5,0.4744 and XD=0.0,0.0,0.0 at the beginning.

Basically, I initialize Moordyn by using LinesInit(X, XD) and then I call LinesCalc(X, XD, Flines, &tprev, &deltaT) function for force calculation.

In the first iteration, I suppose to get zero force along the x and y directions (Fx, Fy) from MoorDyn but I get some forces. (In the first iteration X and XD are the same as at the beginning). So what would be the reason that in the first iteration, I get the wrong force result from MoorDyn?

Thanks in advance.

[lines.txt](https://github.com/FloatingArrayDesign/MoorDyn/files/12358280/lines.txt)

sanguinariojoe:
My two cents, do not use MoorDyn v1

RyanDavies19:
Hi, MoorDyn v1 is the older version of the code, v2 is the most up to date version. v1 should still work for your case though, as it is only lines and connections. Even though your points aren't moving MoorDyn will calculate the forces from the line sinking and from the weight of the lines. Likely the forces you see on the fairlead and anchor points are coming from these line tensions. Does that make sense for the order of magnitude of what you are seeing?

sanguinariojoe:
Many errors during the initialization has been fixed since version 1, not
to mention the memory errors which make the things quite random sometimes.

Would you try the version 2 and let us know?

soydanahmet:
Now I'm able to get reasonable forces with version 2 only if I moved the center of the body to 0,0,0 and create the input file accordingly. If I don't move it, the forces remain as before.

But now, I have another confusion and would very much appreciate it if you can clarify this:

In the method below, X stores the center of mass and the rotation angles of the body. As far as I understand, the unit of rotation angles should be rad/s and the first angle X[3] should be the roll angle, the second  X[4 ] pitch, and the third one X[5] yaw angle. Am I right?

MoorDyn_Step(moordyn_, X, XD, Flines, &tprev, &deltaT);

Thank you.

RyanDavies19:
[...]

PR #109, opened 2023-08-18
Body orientation input to deg

RyanDavies19:
Adjusting body orientation inputs to be degrees per the conversation in issue #108

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -71,7 +71,8 @@ Body::setup(int number_in,
 		bodyM = M_in;
 		bodyV = V_in;
 
-		body_r6 = r6_in;
+		body_r6.head<3>() = r6_in.head<3>();
+		body_r6.tail<3>() = deg2rad * r6_in.tail<3>();
 		body_rCG = rCG_in;
 		bodyI = I_in;
 		bodyCdA = CdA_in;
--- a/source/Misc.hpp
+++ b/source/Misc.hpp
@@ -845,6 +845,9 @@ const real pi = 3.141592653589793238462643383279502884197169399375105820974944;
 /// Constant to convert radians into degrees
 const real rad2deg = 180.0 / pi;
 
+/// Constant to convert degrees into radians
+const real deg2rad = pi / 180.0;
+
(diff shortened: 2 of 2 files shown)


sanguinariojoe:
Some comments (sorry, i do not have a computer to properly check this out):

- This is not backward compatible. Matt has already approved this, so I
suppose it does not matter much.
- Is this not affecting the tests? They should be changed accordingly,
right?
- I recommend making deg2rad and rad2deg inline constant functions instead
of numbers. Just a suggestion though.

RyanDavies19:
@sanguinariojoe When you say backwards compatible are you referring to version 1? I'll change things to inline constant functions.

AlexWKinley:
My guess is that backward comparability in this case means that anyone with an existing Moordyn v2 model that specified the rotation of a body, will have to go in and convert that model to specify as degrees or else get incorrect results when they update MoorDyn.

sanguinariojoe:
Ryan, you pushed [this commit](https://github.com/FloatingArrayDesign/MoorDyn/pull/109/commits/4fe4ac90ae23871878ebaf18ef42b9f0b883a1f0) which has nothing to do with this PR.

On top of that, you are uploading Visual Studio stuff. Please, revert the commit and push (forcing):

git reset edf588685047d6d59f2c8364fe86ce841f65a2a6
git push -f origin v2_changes

Then please add an entry to the .gitignore file to ignore everything related with Visual Studio (.vscode folder), move to another branch and do the commit.

I will merge this as soon as you do that. I will eventually change the tests myself afterwards

RyanDavies19:
@sanguinariojoe Done and changes have been moved to a different branch. Sorry about that, I had forgotten this hadn't been merged yet. Thanks for the help.

RyanDavies19:
> My guess is that backward comparability in this case means that anyone with an existing Moordyn v2 model that specified the rotation of a body, will have to go in and convert that model to specify as degrees or else get incorrect results when they update MoorDyn.

@AlexWKinley my guess is most people have already done this, because prior to this commit inputting degrees would have led to really strange results because the rotations were read by MD as radians.

sanguinariojoe:
Merging right away!

PR #110, opened 2023-08-23
Bug fix for line fluid inertia forces, and Wavekin 7 Ud_z

AlexWKinley:
When lines compute the fluid inertia forces, they didn't include segment submergence in the calculation. It's possible that the code as written expects the volume array to take that into account, which does not currently happen. If you look at the section that computes the mass matrix for each node, it includes a correction for submergence along with the volume array. If we want to rewrite these so that the volume array is actually submerged volume, that's definitely possible. I don't know if there's an advantage one way or the other. 

Additionally in the code for wavekin 7, which sums up component waves at all the nodes, there was a missing minus sign for the z component of the water acceleration. It looks like this issue is specific to this wave mode, it is not present in wave mode 2, which uses the same function to calculate those values as wave mode 3, so those probably both work correctly. 

At some point I'll probably write some water velocity/acceleration unit tests for all the wave modes just to sanity check them, and make it easy to validate any future wave modes.

Update:

I've also gone ahead and fixed a relative minor issue with the wave kinematics (that I introduced along with wave stretching), where previously points out of the water were given zero velocity, and acceleration for their wave kinematics. This can cause issues in situations where a point is near the water surface, and the object itself has some amount of submergence. A good example would be a line modeling some sort of floating hose, what can float with its central axis out of the water. Now, points out of the water get the water kinematics at the waters surface below them, and have to do their own handling of submergence.

### DIFF
--- a/source/Waves/SpectrumKin.cpp
+++ b/source/Waves/SpectrumKin.cpp
@@ -57,25 +57,22 @@ SpectrumKin::getWaveKin(vec3 pos,
 	    amplitudes * (omegas * t - kValues * distances + phases).sin();
 	const Eigen::ArrayX<real> cos_waves =
 	    amplitudes * (omegas * t - kValues * distances + phases).cos();
-	const real surface_height = sin_waves.sum();
 
-	if (pos.z() > surface_height) {
-		if (zeta) {
-			*zeta = surface_height;
-		}
-		if (vel) {
-			*vel = vec3::Zero();
-		}
-		if (acc) {
-			*acc = vec3::Zero();
-		}
-		return;
-	}
+	const real surface_height = sin_waves.sum();
 	const real bottom = actualDepth;
 	const real actual_depth = surface_height - bottom;
 
-	const real stretched_z =
+	real stretched_z =
 	    (-avgDepth * (pos.z() - bottom)) / actual_depth + avgDepth;
+
+	// If the point is above the water surface, return the values at the water
+	// surface. This is important for situations where a point is out of the
+	// water but the object itself may have significant submergence (sideway
+	// floating line, top node of buoy, etc)
+	if (stretched_z > 0.0) {
+		stretched_z = 0.0;
+	}
+
 	if (zeta) {
 		*zeta = surface_height;
 	}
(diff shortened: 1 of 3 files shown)


source/Waves/SpectrumKin.cpp
```diff
@@ -57,25 +57,22 @@ SpectrumKin::getWaveKin(vec3 pos,
...
+	const real surface_height = sin_waves.sum();
 	const real bottom = actualDepth;
 	const real actual_depth = surface_height - bottom;
 
-	const real stretched_z =
+	real stretched_z =
 	    (-avgDepth * (pos.z() - bottom)) / actual_depth + avgDepth;
+
+	// If the point is above the water surface, return the values at the water
+	// surface. This is important for situations where a point is out of the
```
mattEhall, line 69: Why is this important (or correct)?

mattEhall:
Most of this looks great. Good catch on adding the submergence fraction for fluid inertial loads on lines.

I don't think there was actually an issue with the wave stretching before. At least from a quick look, it seems like Wheeler stretching is being used, but the adjustment in this PR caps things at the free surface in the manner of "vertical stretching", which is a different method. I think not doing this capping was correct. Does it cause any problems?

source/Waves/SpectrumKin.cpp
AlexWKinley: You can have a situation like a line floating on the water where the nodes around a segment may be above the water surface (by a small amount), but the segment still has a significant submergence, then surface currents will create drag on the segment in an amount proportional to the submergence and surface water velocity.

If the line nodes are told there is zero water velocity, then there will be no drag forces applied to that segment. 
If the line nodes are told the velocity at the water surface, then the segment will have drag forces calculated in a way that at least approximates reality because lines account for submergence in their drag calculations.

AlexWKinley:
With Wheeler stretching, a "stretched z" of 0 always refers to the instantaneous free surface. So if we first stretch, and then cap the stretched z to a value of 0, we always return the value at the free surface. And if we don't add a cap, we could end up returning values where the vertical scaling factor (like cosh(k * (z + h)) / sinh(kh)) is increasing the velocities and acceleration in a way that doesn't reflect what the object partially submerged would experience. 

I might agree that in most cases this actually has a negligible effect, the thing with the bigger effect is removing the pos.z > surface_height check that would return zero velocity. If you think it's more correct without the stretched z capping, it's definitely an easy change.

sanguinariojoe:
@mattEhall would you change the stretched z capping or shall I proceed merging?

sanguinariojoe:
I merge this before https://github.com/FloatingArrayDesign/MoorDyn/issues/112 happens. Later on we can decide whether we go with the stretched z capping or not

PR #117, opened 2023-09-08
Big docs update

RyanDavies19:
Finished the documentation to explain usage of both MoorDyn-C and MoorDyn-F (in OpenFAST). Updated documentation organization and finished all incomplete sections. New version includes a much more updated input file section, with explanation of all the output flag options. 

When I was compiling these docs locally the inline C_API references were not working. I believe though this had something to do with how I set up my system locally, because the old docs were giving the same errors. When these are compiled here we will need to check that they are working.


sanguinariojoe:
Hey Ryan!

From this point onwards the PRs shall be directed towards "dev" branch.

master should be some sort of stable branch, so we just merge there things
from dev time to time.

Remember that pushing things on master triggers the creation of a new
release with the artifacts crafting, as well as python packaging

RyanDavies19:
@sanguinariojoe Thanks for the clarification there. Changed to dev

PR #147, opened 2023-10-05
Manylinux Python wheels

sanguinariojoe:
Alright! This is finally working!

https://test.pypi.org/project/moordyn/2.2.2/#files

I excluded, for the time being, the musllinux platforms, which is popular just on embedded devices. They can anyway install moordyn using the sdist, so they are actually supported by the same pip command


sanguinariojoe:
Uhm... problems with the regular buildings... Weird...

sanguinariojoe:
It works now!

PR #164, opened 2023-10-17
Adjustments to Dynamic relaxation

### LINKED ISSUE #160: Static line tension disagreement
Static lines initialize with different tensions between MDC_v2 and MDF_v2/MDC_v1. Differences also emerge with dynamic relaxation. Using [MoorPy](https://github.com/NREL/MoorPy), NREL's quasi-static model, the equilibrium fairlead tension is 4808 kN. This indicates that MDC_v1 and MDF_v2 are getting a solution closer to the correct answer. See the figures below for a single line resting partially on the seabed:

 Line profile:
case_1_shape

Fairlead tension with TmaxIC = 0:
case1_comparison

Fairlead tension with TmaxIC = 200:
200s_dyrelax

RyanDavies19:
Resolves #160 

This PR removes the double counting of line weight and changes the dynamic relaxation routine to match the now fixed fortran code. The changes to the dynamic relaxation routine are as follows:
- The user inputted dtIC is subdivided into 10 steps, with the intent that the dtIC is the time that the user wants to see stability over for convergence
- Changes number of convergence iterations so that 9 iterations are compared to the current fairten (because of subdividing each dtIC into 10 pieces)
- Changes the while loop parameter to a tolerance (this can be adjusted, just seemed like the easiest method) which eliminates an extra timestep being run when convergence is not reached. Prior to this, the final time after the loop would be ICTmax+dtIC, now it is just ICTmax
- Changed iic indexing to match the indexing in fortran so the printed outputs are the same

### DIFF
--- a/source/Line.hpp
+++ b/source/Line.hpp
@@ -429,9 +429,11 @@ class Line final : public io::IO
 			throw moordyn::invalid_value_error("Invalid node index");
 		}
 		if ((i == 0) || (i == N))
+			// return (
+			//     Fnet[i] +
+			//     vec(0.0, 0.0, M[i](0, 0) * (-env->g))); // <<< update to use W
 			return (
-			    Fnet[i] +
-			    vec(0.0, 0.0, M[i](0, 0) * (-env->g))); // <<< update to use W
+			    Fnet[i]); // <<< update to use W
 
 		// take average of tension in adjacent segments
 		return (0.5 * (T[i] + T[i - 1]));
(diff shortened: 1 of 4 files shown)


RyanDavies19:
Not sure why pendulum is failing, all looks fine on my end. Will dig into this tomorrow.

sanguinariojoe:
> Not sure why pendulum is failing, all looks fine on my end. Will dig into this tomorrow.

If you set the error tolerance to 2.5% it works (line 150):

`CHECK_VALUE("T", t_peaks[i], i * 0.5 * T, 0.025 * T, t_peaks[i]);`

Not a big deal I suppose

PR #166, opened 2023-10-23
Clarifies some parts of the documentation

RyanDavies19:
These updates hopefully help clear up some confusion seen in issues and messages I have received about using the step function.

### DIFF
--- a/docs/compiling.rst
+++ b/docs/compiling.rst
@@ -16,48 +16,42 @@ OpenFAST/Fast.Farm or with its own driver), please refer to the
 MoorDyn-F in a standalone form can be compiled as moordyn_driver inside of the build folder.
 
 MoorDyn-C is available on `GitHub <https://github.com/FloatingArrayDesign/moordyn/>`_. The 
-latest version can be found in the V2 branch. MoorDyn-C v1 is currently in the master branch
+latest release can be found in the master branch. MoorDyn-C v1 is currently in the master branch
 of this repo, and it can be compiled using these :ref:`instructions <compile_simple>`. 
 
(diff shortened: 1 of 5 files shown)


RyanDavies19:
@sanguinariojoe thanks getting that merged. My intent with merging to the master branch was because these all were documentation fixes (typos, clarifications, etc.) that referred to the current code rather than to any new features. I understand it is not best practice to PR directly to the master branch. Is there some way we could update the docs with those changes? 

There are also a number of future docs improvements I am going to add based on feedback I've gotten. So it would be good to have a process for improving the docs, or at least pointing people to the most up to date version.

sanguinariojoe:
I think frequent releases is more convenient.

Committing to master imply a lot of CI/CD automatic operations... Crafting
releases, publishing python packages...

I suggest to define a protocol to make "emergency" releases (you know, to
fix critical bugs or to address heavy flaws on the docs)

sanguinariojoe:
To be more specific, I suggest to define a protocol in which a PR is merged
on dev, immediately followed by the PR against master (which triggers the
next release). Some kind of shortcut to avoid waiting for the rest of
planned features.

Maybe a special label? A prefix on the PR title?
And maybe a list of privileged devs which might ask for that declared on
the documentation/readme? Sort of a steering committee?

RyanDavies19:
@sanguinariojoe That sounds like a good plan. At least for now I think just managing it on a case by case basis will work because there is not a ton of external contribution requests. As for the documentation I can just adjust so that the main page of the docs points to the dev branch, which should work for docs patches.

Issue #163, opened 2023-10-16
Euler angle conventions and Body input/output angles


hamzaJayy:
Hi all, 
I installed MoorDynv2 matlab wrapper. I was testing input and output angles for Body object. 
[lines_test.txt](https://github.com/FloatingArrayDesign/MoorDyn/files/12912904/lines_test.txt)

I am modelling a single Body as coupled object and trying to provide motion externally calculated. 

the script I am using is below:
clc;close all;clear
%% Create MoorDyn system
 system1 = MoorDynM_Create('lines_test.txt');
%% find num of DOF
DOF =MoorDynM_NCoupledDOF(system1);
x = zeros(1,DOF);
dx = zeros(1,DOF);

%% Setup the initial condition
MoorDynM_Init(system1, x, dx);

%% Simulate
dt =0.04;
t = 0:dt:10;
x =zeros(length(t),DOF);
dx =zeros(length(t),DOF);
% provide roll angle ansd speed
roll_speed = 5; % deg/sec
x(:,4) = t*roll_speed;
dx(:,4) = roll_speed;
% run the moordyn
tic
for i=1:length(t)-1
[sim_time, f] = MoorDynM_Step(system1, x(i,:), dx(i,:), t(i),dt);
end
%% Alright, time to finish!
MoorDynM_Close(system1);
toc
%% plot
data=readtable('lines_test_Body1.out','FileType','text');
VariableNames = {'Surge','Sway','Heave','Roll','Pitch','yaw'};
indx =4;
figure(1);plot(t,x(:,indx))
hold on
plot(t,data{:,1+indx})
hold off
title(VariableNames{indx})

I am only giving 5 deg/s angular speed in roll. Since there is no other element, i was expecting the roll angle will be same for input and ouput. Not only the roll angles saved by MoorDyn is quite different than expected, but unexplained pitch and yaw angles. 

![image](https://github.com/FloatingArrayDesign/MoorDyn/assets/106804145/bd1d6024-905c-49cc-9d3b-4ef4353cd150)

![image](https://github.com/FloatingArrayDesign/MoorDyn/assets/106804145/bfeda13a-4fb2-4693-9a9a-d360abe82fdc)

![image](https://github.com/FloatingArrayDesign/MoorDyn/assets/106804145/6db92ea1-cd2a-4d13-be01-f78d7b0246e2)

Can anyone explain whats going on here ? Thanks in advance.

AlexWKinley:
[...]

PR #167, opened 2023-10-26
Added canonical Euler angles for outputs

RyanDavies19:
Based on discussion in #163 

Adds canonical Euler angles function from Eigen to replace the Euler angle function. This resolves an issue where outputted roll could not be negative, and where certain input body rotations would not be reflected in the output rotation values. This only changes the values written to output files, not internal dynamics. 

Below are outputs before and after the change for a coupled body rotating periodically.

Before:
outputs before change

After:
after_adjustment

@AlexWKinley thanks for putting together the suggested code for this, I only had to make some minor changes. 

We can move this into Misc.cpp if we feel it goes better there. I put it where it is because that's where the Quat2Euler functions are.

### DIFF
--- a/source/Misc.hpp
+++ b/source/Misc.hpp
@@ -142,10 +142,93 @@ typedef ivec3 ivec;
 /// COmplex numbers
 typedef std::complex<real> complex;
 
+inline vec3 
+canonicalEulerAngles(const quaternion& quat, int a0, int a1, int a2)
+{
+	// From issue #163: https://github.com/FloatingArrayDesign/MoorDyn/issues/163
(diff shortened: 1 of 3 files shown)


RyanDavies19:
Update: I changed the body test so that it now passes. Based on the print statements the body needs to have a quat along the x axis with 90 angle. Before canonical angles, this was an axis in the positive x direction with a -90 deg angle. Now the axis is in the negative x direction with a +90 deg angle. Because they are equivalent then the test should be considered passing. I also compared to MDF with an attached rod and the behavior/orentation remains the exact same as it did before canonical Euler angles

PR #179, opened 2024-01-02
Pinned bodies

RyanDavies19:
Second try of #178. Everything should be rebased and up to date now.

This PR includes a new option for bodies, allowing them to behave as a pinned object. It contains a corresponding unit test and documentation updates. It also contains a feature update that prints time steps to the terminal while running, similar to how MDF works at runtime.

The formulation follows the pinned rods orientation closely. Inertial responses and forces have been added to the getStateDeriv and getFnet functions for both bodies and rods. The torque from translational acceleration is:
F_iner [4:6] = M.bottomLeftCorner * a6[1:3]

and the inertial response force to acceleration is given as (contains two terms to avoid counting the externally applied force):
F_iner[1:3] = - M.topLeftCorner * a6[1:3] - M.topRightCorner * a6[3:6]

This feature has been verified in MDF using CAD software, analytical solutions, and Subdyn. For the MDC version I verified using MDF, and the two versions match very closely.

The corresponding update will be included in OpenFAST v4.

This update did require the addition of an acceleration term for each calling time step for the calculation of inertial forces. In MDF acceleration a value passed into the step function. For MDC, I did this the same way velocity is handled: constant over the coupling step. While this isn't the most accurate, it was the simplest approach and in most use cases the coupling time step will be small enough that the effect is negligible (there was never issues when comparing with MDF).


RyanDavies19:
@sanguinariojoe I was not able to reopen #178 so I just made a new PR.

PR #194, opened 2024-02-14
Add the possibility to set the constant EA on real time

sanguinariojoe:
For polyester lines it might be required to set the EA, as well as the line length, depending on the mean value of the tension.

In this PR we add an entry on the API to set such EA value, as well as a polyester line test


RyanDavies19:
@sanguinariojoe Does this then allow for a user to set the line EA between each call to MD_step? And if so, can the user still just provide an input file for nonlinear EA as previously it was set up? 

Also apologies for letting the time integrator PR sit, I have been wrapped up in some other work that has taken up most of my time. I hope to get to it in the near future, probably march

sanguinariojoe:
> @sanguinariojoe Does this then allow for a user to set the line EA between each call to MD_step?

Yes

> And if so, can the user still just provide an input file for nonlinear EA as previously it was set up?

Nope, right now if the line has a non-linear EA all those new API entries are useless. We can though add some more entries to the API.

If we go this way, I rather add something to can set the raw values instead of an input file. What do you think?

> Also apologies for letting the time integrator PR sit, I have been wrapped up in some other work that has taken up most of my time. I hope to get to it in the near future, probably march

No prob. It is good just to know you are planning to address it

sanguinariojoe:
I am merging this because it is anyway harmless

RyanDavies19:
@sanguinariojoe Thanks for merging, this looks good. I talked with Matt and there is a similar feature in progress for MD-F so its good to keep them consistent with this. I'll note just for later reference that I checked and providing a filename with a lookup table instead of a value for the EA is still supported (as defined in the docs).

PR #201, opened 2024-04-02
Fixes #196

### LINKED ISSUE #196: Unnecessary point in input file for lines using only rods
I have a MoorDyn input-file defining a cable with bending stiffness: [exampleinput.txt](https://github.com/FloatingArrayDesign/MoorDyn/files/14508108/exampleinput.txt)
To define the cable, I only use rods as connections. It would not need any points. If I delete the points-section from the input-file, I obtain the following error, and MoorDyn does not run:

`ERR source\MoorDyn2.cpp:952 moordyn::MoorDyn::ReadInFile(): Reading lines without defined points`
`Error (-3) during the Mooring System creation:`
`Exception while reading the input file`

Therefore, I defined a point in the input-file which is actually not needed in the simulation.
I am using MoorDyn v2.0.0, 2023-09-18 and v2.2.2 on Windows 10.

sanguinariojoe:
Even though the issue was closed,  @anjaschnepf made a point there. Lines can be defined without points, so we shall not trigger an error

### DIFF
--- a/source/MoorDyn2.cpp
+++ b/source/MoorDyn2.cpp
@@ -1009,10 +1009,6 @@ moordyn::MoorDyn::ReadInFile()
 			LOGERR << "Reading lines without defined line types" << endl;
 			return MOORDYN_INVALID_INPUT;
 		}
-		if (!PointList.size()) {
-			LOGERR << "Reading lines without defined points" << endl;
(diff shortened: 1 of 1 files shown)


RyanDavies19:
Thanks for fixing this! I will merge

RyanDavies19 (approved):

PR #206, opened 2024-04-15
Prevent body drag coeffs from being negative

### LINKED ISSUE #205: Body negative local drag coefficients
Strictly speaking, this is a bug that I encountered on our internal MoorDyn fork, which is somewhat out of date. I don't see anything in the latest public code that would address this, but I haven't directly tested this behavior with the latest branch of this repo.

When bodies are in certain orientations, the computed global drag coefficients can end up having negative values.

The code at fault is
https://github.com/FloatingArrayDesign/MoorDyn/blob/99cd2b6e6afa7e5d58e2ac7ac3a4b05e127f1f72/source/Body.cpp#L496-L501

An example of a case where this fails to work is when the body's pitch is -90 degrees. The rotation matrix in this case will be
```
[  0.0,  0.0, -1.0;
   0.0,  1.0,  0.0;
   1.0,  0.0,  0.0 ]
```

The `cda` variable will look like `bodyCda * [1, 1, -1, 1, 1, -1]`. Which leads to a negative drag coefficient in the z axis, which is a destabilizing force that is both not physical and will eventually cause the simulation to error out.

I'm making this issue because I'm not sure what the correct solution to this issue is. I think it's valid to just take the absolute value of the rotated drag coefficients, but I'll defer to others for how they want to address this.

AlexWKinley:
fixes #205 

I think this should be a sufficient change to address the issue of getting negative drag coefficients in certain orientations.

The new body test fails without this change because its z drag coefficient is negative, causing the body to fly off the infinity. With the fix the test passes normally. Once others are happy with this change, I'll fix the merge conflict.

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -581,7 +581,7 @@ Body::doRHS()
 	cda(Eigen::seqN(0, 3)) = OrMat.transpose() * bodyCdA.head<3>();
 	cda(Eigen::seqN(3, 3)) = OrMat.transpose() * bodyCdA.tail<3>();
 	F6net +=
-	    0.5 * env->rho_w * vi.cwiseProduct(vi.cwiseAbs()).cwiseProduct(cda);
+	    0.5 * env->rho_w * vi.cwiseProduct((vi.cwiseProduct(cda)).cwiseAbs());
 
 	// Get contributions from any points attached to the body
 	for (auto attached : attachedP) {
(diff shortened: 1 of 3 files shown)


RyanDavies19:
@AlexWKinley I have talked with Matt about this and I think there is a better approach. The translational and rotational drag coefficients should each be converted to a 3x3 matrix, and then multiplied by the orientation matrix and it's transform. This should yield a 3x3 drag matrix that then is multiplied by the velocity to give our drag force. We would do this for both the translational and rotational drag. coefficients, giving us an end result of a 6 DOF drag force vector:

```math
F \\
=
0.5 * \rho_w * \\
OrMat\\
\begin{bmatrix}
CdA_x & 0 & 0 \\
0 & CdA_y & 0 \\
0 & 0 & CdA_z  \\
\end{bmatrix}
OrMat^T *\\
(\vec v_i  * \\
|\vec v_i |)
```

I'll put together a PR for this and have it up soon.

PR #209, opened 2024-04-18
Minor fix in velocities for body drag calc

RyanDavies19:
See conversation in #208. Fixing a minor bug in how velocities are multiplied for body drag coefficients. 

@AlexWKinley does this look like what you had in mind?


AlexWKinley:
Looks good to me!

PR #210, opened 2024-05-09
Give back the same values by the output channels

sanguinariojoe:
This should fix the problem found by @RyanDavies19 at https://github.com/FloatingArrayDesign/MoorDyn/pull/202

The API still works the new way though.

As I suppose it can be appreciated, the former way is quite inconsistent, at least with respect the fairlead and the anchor. Maybe it would be good to take the axial damping out to it own API method (`MoorDyn_GetLineNodeTenD()` and `moordyn::Line::getNodeTenD()`)

### DIFF
--- a/source/Line.cpp
+++ b/source/Line.cpp
@@ -531,12 +531,15 @@ Line::GetLineOutput(OutChanProps outChan)
 		return rd[outChan.NodeID][1];
 	else if (outChan.QType == VelZ)
 		return rd[outChan.NodeID][2];
-	else if (outChan.QType == Ten)
-		return getNodeTen(outChan.NodeID).norm();
+	else if (outChan.QType == Ten) {
+		if ((outChan.NodeID == 0) || (outChan.NodeID == N))
+			return getNodeForce(outChan.NodeID).norm();
+		return (0.5 * (T[outChan.NodeID] + T[outChan.NodeID - 1])).norm();
+	}
 	else if (outChan.QType == TenA)
-		return getNodeTen(0).norm();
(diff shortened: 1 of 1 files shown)


source/Line.cpp
```diff
@@ -531,12 +531,15 @@ Line::GetLineOutput(OutChanProps outChan)
 		return rd[outChan.NodeID][1];
 	else if (outChan.QType == VelZ)
 		return rd[outChan.NodeID][2];
-	else if (outChan.QType == Ten)
-		return getNodeTen(outChan.NodeID).norm();
+	else if (outChan.QType == Ten) {
+		if ((outChan.NodeID == 0) || (outChan.NodeID == N))
+			return getNodeForce(outChan.NodeID).norm();
+		return (0.5 * (T[outChan.NodeID] + T[outChan.NodeID - 1])).norm();
```
RyanDavies19, line 537: I think here for internal nodes we should return the wall tension using `Line::getNodeTen(outChan.NodeID).norm()`. This would keep output channels consistent with what is done in MDF, where the subroutine GetLineTen is returned for the "Ten" output channel: https://github.com/OpenFAST/openfast/blob/6a7a543790f3cad4a65b87242a619ac5b34b4c0f/modules/moordyn/src/MoorDyn_IO.f90#L1885

sanguinariojoe:
Oki, tomorrow I will try to do that

RyanDavies19:
@sanguinariojoe Thanks for making those fixes, all looks good compared to MDF now so I will merge.

Figure_3

Figure from the same test case mentioned in #202

PR #218, opened 2024-05-29
Bending due to pressure forces

sanguinariojoe:
This effect would be important for pipes, like risers, due to both the external pressure and the internal flow.

Some notes:

 - It is disable by default, so it is harmless.
 - It is not controllable from the config file, just from the API.
 - To enable it the C API function MoorDyn_SetLinePressBend() shall be called
 - Internal flow pressure can be set with the C API entry MoorDyn_SetLinePressInt()
 - ~The wrappers are not ready yet. I hope I will have them ready today or tomorrow~

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -560,7 +560,7 @@ Body::doRHS()
 
 	// env->waves->getU(r6, t, U); // call generic function to get water
 	// velocities <<<<<<<<< all needs updating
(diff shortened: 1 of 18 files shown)


sanguinariojoe:
Wrappers done!

RyanDavies19:
@sanguinariojoe looks good on my end, double checked bending stiffness calcs with MDF and still lining up exactly.

RyanDavies19 (approved):

PR #223, opened 2024-06-18
Docs update: typos, more references, and updated FF file

RyanDavies19:
Fixed typos found and/or reported by users. Aims to clear up common questions seen on OepnFAST help pages. Updates the FAST.Farm example to be consistent with current capabilities. Adds more references to publications about feature improvements.

### DIFF
--- a/source/MoorDyn2.cpp
+++ b/source/MoorDyn2.cpp
@@ -108,11 +108,11 @@ moordyn::MoorDyn::MoorDyn(const char* infilename, int log_level)
 		_basepath = _filepath.substr(0, lastSlash + 1);
 	}
 
-	LOGMSG << "\n Running MoorDyn (v2.0.0, 2023-09-18)" << endl
+	LOGMSG << "\n Running MoorDyn (v2.3.1, 2024-5-24)" << endl // TODO: can we automate this?
 	       << "         MoorDyn v2 has significant ongoing input file changes "
 	          "from v1."
 	       << endl
-	       << "   Copyright: (C) 2023 National Renewable Energy Laboratory, "
+	       << "   Copyright: (C) 2024 National Renewable Energy Laboratory, "
 	          "(C) 2014-2019 Matt Hall"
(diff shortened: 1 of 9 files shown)


sanguinariojoe:
Hey! This PR will introduce some conflicts on https://github.com/FloatingArrayDesign/MoorDyn/pull/182.

I suggest the following:

- I finish the MDF automated verification
- We check if https://github.com/FloatingArrayDesign/MoorDyn/pull/182 is verifying OK (I would expect some differences due to the initialization, but so far the same results).
- If https://github.com/FloatingArrayDesign/MoorDyn/pull/182 is OK, we merge it and fix this. Otherwise we merge this and fix https://github.com/FloatingArrayDesign/MoorDyn/pull/182

Is that OK?

source/MoorDyn2.cpp
```diff
@@ -108,11 +108,11 @@ moordyn::MoorDyn::MoorDyn(const char* infilename, int log_level)
 		_basepath = _filepath.substr(0, lastSlash + 1);
 	}
 
-	LOGMSG << "\n Running MoorDyn (v2.0.0, 2023-09-18)" << endl
+	LOGMSG << "\n Running MoorDyn (v2.3.1, 2024-5-24)" << endl // TODO: can we automate this?
```
sanguinariojoe, line 111: The version for sure! The date would be quite annoying... However I think we can just drop it, since we have now automated releases with the timestamp. Everyone can check when do we release a new version

RyanDavies19: Agreed, as long as the version is up to date users can reference the release.

RyanDavies19:
@sanguinariojoe that sounds like a good plan. I will have more reg tests up next week for comparison. Let me know if there are any you want to compare.

sanguinariojoe:
I have changed my mind... I am merging this, launching a patch release, and then we can continue with the other tasks (verification and time marching)

PR #229, opened 2024-06-26
Get the MoorDyn version on the legacy compilation systems

sanguinariojoe:
Fixes the regression introduced on #225, which made impossible to compile with the legacy makefiles

@RyanDavies19 , please test it on your Mac and eventually merge. On the other hand, at some point we should address your problems with CMake


RyanDavies19:
Works great on Mac, thanks for putting this together. Cmake works fine for me, I just think its good to keep these legacy methods around because they're fool-proof if users are having compile issues, they just need to run make.

RyanDavies19 (approved):

PR #232, opened 2024-07-03
Update MD-F comparison to point to OpenFAST/dev

RyanDavies19:
OpenFAST/dev now has a more robust suite of regression tests that should capture most of MoorDyn's capabilities including 6 DOF dynamics and cable. Our checking of things should be against the dev branch for the time being until it gets merged into OpenFAST/main. This might break the MD-F tests for #231.

Corresponding OpenFAST PR: https://github.com/OpenFAST/openfast/pull/2294


RyanDavies19:
Consider this in progress for now, I will try to address the failing tests Monday after the long weekend for the 4th of July. Will also try to get GH artifacts going so we can download plots of the failed tests if/when they fail

sanguinariojoe:
OK, I am just waiting until you give me a heads up

RyanDavies19:
Okay we now have test artifacts with plots of channels and error thresholds so we can see what is broken. As for why those tests don't line up, that will take further investigation.

RyanDavies19:
@AlexWKinley and @sanguinariojoe, not sure if either of you have a rush for exact kinematics but what's in MD-C at the moment isn't entirely right. Once https://github.com/OpenFAST/openfast/pull/2334 is merged, a corresponding update needs to be made on the MD-C side to the body and rod kinematics (adding centripetal moments in addition to the centripetal forces from #231). That will be a more correct approach to 6 DOF kinematics. 

After that is done, tests here should agree with the exception of md_lineFail. md_lineFail will be fixed when I get around to updating our line failures

sanguinariojoe:
Everything good on my end, no need to rush. I am busy with some other interesting stuff for MoorDyn! :-p

AlexWKinley:
Similarly, no rush from my end. (Although I suspect that exactly correct kinematics in 6dof are still more complicated than any of our numerical models have accounted for). I'm busy with other work for the moment. I do super appreciate the work that's been happening here in terms of checking agreement between the two MoorDyn implementations.

sanguinariojoe:
> @AlexWKinley and @sanguinariojoe, not sure if either of you have a rush for exact kinematics but what's in MD-C at the moment isn't entirely right. Once [OpenFAST/openfast#2334](https://github.com/OpenFAST/openfast/pull/2334) is merged, a corresponding update needs to be made on the MD-C side to the body and rod kinematics (adding centripetal moments in addition to the centripetal forces from #231). That will be a more correct approach to 6 DOF kinematics.
> 
> After that is done, tests here should agree with the exception of md_lineFail. md_lineFail will be fixed when I get around to updating our line failures

Hey! How are you doing with this? Are you preparing a PR?

RyanDavies19:
Hi @sanguinariojoe, I haven't had much time recently to take a stab at this, and I probably won't for the next couple weeks. I've blocked out some time at the end of October/Early November to work on this. Aside from updating the line failure capability on this side, there are a couple of areas I suspect need to be aligned Namely hydrodynamics of rods, submergence of lines, and rigid body kinematics. 

On a more logistical note, I will try to keep the MD-F regression tests to the core MoorDyn capabilities. That way as new auxiliary features are added on either side, they won't cause the tests the fail while we still maintain the same performance with the core things (6 DOF kinematics, line modeling, etc.)

sanguinariojoe:
OK, as long as you have it planned I just wait

RyanDavies19:
I am closing this for now, given it has gone a bit stale. Eventually we will need an effort to bring the two codes in line with each other and from then on we should have this tests run for all PR's

PR #265, opened 2024-10-24
MATLAB wrapper testing

sanguinariojoe:
A GitHUB action to compile the MATLAB mex files and run a small test. Notes:

- It is quite shitty because it can not be integrated with CMake
- I tried to make it work on Windows, but it gives an error with little to none useful information to fix it.
- At some point it would be great to have a common framework to create and run the tests with both GitHUB actions and CMake

### DIFF
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -1,13 +1,16 @@
 name: Test MATLAB MEX
 
 on:
-    branches: [ $default-branch, master, wrappers_test ]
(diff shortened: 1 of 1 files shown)


sanguinariojoe:
PS I hate you MATLAB

RyanDavies19:
@sanguinariojoe this looks good for me, thanks for adding. Annoying it won't work for CMake but just a simple check like this for now is fine.

PR #267, opened 2024-11-05
fix(core): Water depth warning was printed no matter whether it was corrected or not

sanguinariojoe:
Just a tiny fix


RyanDavies19:
@sanguinariojoe good catch, I also had this as a fix in some things I've been working on

PR #269, opened 2024-11-06
State and Line Performance Improvements

AlexWKinley:
This PR contains changes to a few time integrator and to lines with the goal of not changing the simulation is any meaningful way, while providing performance improvements.

# Performance Results

Overall, for models of lines, there is roughly a 2x performance improvement. 
Currently, the integrator performance improvements are mostly specific to RK2 and RK4.

RK2
RK4

line_performance_plots

# Integrator / State Changes

The primary changes to the integrator and state code are to avoid memory allocations.

The state code that allows writing
```cpp
r[1] = r[0] + rd[0] * (0.5 * dt);
```
is very convenient, but its current implementation means that every operation on a state allocates memory to create an entire copy of that state. That basic $y_{n+1} = y_n + dy * \Delta t$ line allocates memory to store the result of $dy * \Delta t$, and then the result of the addition, and even an additional allocation for the assignment. 

My current solution to this is the new `butcher_row` function that can do these sorts of computations in place, without allocation any additional memory. Named as such because it does the math for a single row in a [Butcher tableau](https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods#Explicit_Runge.E2.80.93Kutta_methods).
With this function we get
```cpp
// r[1] = r[0] + rd[0] * (0.5 * dt);
butcher_row<1>(r[1], r[0], { 0.5 * dt }, { &rd[0] });
```
Definitely not a clear and concise as the more mathematical notation, but with significant performance improvements.
Because of the additional complexity, I've only switch the RK2 and RK4 integrator to use this function. But I can certainly expand its usage if desired.

The other change to states I've made is to remove the empty constructor and destructor from StateVar and StateVarDeriv. Because of the [rule of three/five](https://en.cppreference.com/w/cpp/language/rule_of_three) defining a destructor (even an empty one), causes an expression like `r[1] = r[0] + rd[0]` to allocate twice, even though the result of `r[0] + rd[0]` can be directly assigned to `r[1]`. Removing the empty destructor allows the compiler to avoid that second allocation. 

This should help other integrators even if they're not using butcher_row, but not as significantly as using butcher_row.

# Line Changes

The first type of changes to lines are again to avoid memory allocations. 

Changing `line->getStateDeriv()` to take references to the node velocities and accelerations avoids having to allocate memory for those results every time. 
Also changing `Line::setState(const std::vector<vec>& pos, const std::vector<vec>& vel)` to take vector references avoids it allocation copies of the position and velocities vectors.

The other kind of change to lines are my attempts at simplifying the code while improving performance.
The biggest thing is getting rid of many of the 
```cpp
if (i == 0) 
    ....
else if (i == N)
   ....
else
   ....
```
by calculating the values that depend on whether it's an internal or external node once, allowing for them to be reused, and making the code nicer to read.

# Misc

To improve consistency of the benchmarks, and avoid filling up the console with simulation times, I added `MoorDyn::SetDisableOutput` to disable some of the console and file output. 

# Logistical Notes

Feel free to share any thoughts/questions you have.
I know that those at NREL are doing some work from their side that could potentially create some conflicts with some of these changes. I'm happy to delay merging this and rebasing on top of whatever those changes may be myself if that would be easiest.


AlexWKinley:
@RyanDavies19 
Just a friendly ping in case you haven't seen this.
I'm in no particular rush for this to be merged, but I do want to make sure that if I'm creating any conflicts with work you're doing, we coordinate on making sure those can be resolved.

source/State.hpp
```diff
@@ -442,4 +418,74 @@ class DMoorDynStateDt
...
+			for (unsigned int j = 0; j < N; j++) {
+				line.pos[i] += scales[j] * derivs[j]->lines[lineIdx].vel[i];
+				line.vel[i] += scales[j] * derivs[j]->lines[lineIdx].acc[i];
+			}
+		}
+	}
+
+	for (unsigned int pointIdx = 0; pointIdx < out_state.points.size();
+	     pointIdx++) {
+		auto& conn = out_state.points[pointIdx];
```
RyanDavies19, line 461: To keep things consistent can we change conn to point?

source/MoorDyn2.hpp
```diff
@@ -630,6 +637,9 @@ class MoorDyn final : public io::IO
 	/// (if using env.WaveKin=1)
 	unsigned int npW;
 
+	/// Disabled writing to output files or console when running
+	bool disableOutput = false;
```
RyanDavies19, line 641: We should add this to the input file options list and the docs. I can open a PR to your branch @AlexWKinley.

source/State.hpp
```diff
@@ -442,4 +418,74 @@ class DMoorDynStateDt
...
+ *
+ * @tparam N Number of columns in the row
+ * @param out_state Where to save the new state
+ * @param start_state Starting state
+ * @param scales Derivative weights, one for each derivative state
+ * @param derivs State derivative values
+ */
+template<unsigned int N>
+constexpr void
+butcher_row(MoorDynState& out_state,
```
RyanDavies19, line 439: I like this approach and the benefits we get. In-terms of work I doing, I can move my changes into this format quite easily. I have two additional states that have been added to lines, and the hackish way I've done it for now is to create another state vector that is a copy of the line state class. At the start of State,hpp:
```
class MoorDynState
{
  public:
	/// @brief Constructor
	MoorDynState() {}

	/// @brief Destructor
	~MoorDynState() {}

	/// The states of the lines
	std::vector<LineState> lines;

	/// The state derivatives of misc states: VIV, viscoelastic, blank (acc is dummy variable)
	std::vector<LineState> misc;
```

I realize this it not very efficient, but for timelines for the project that needed this work it was the fastest approach. When I open a PR with this we can discuss more.

source/MoorDyn2.hpp
RyanDavies19: PR is open and ready to merge. It adds this, and also cleans up the options list handling a bit

AlexWKinley: Merged. Thanks for cleaning that up.

source/State.hpp
AlexWKinley: I agree that seems a little hacky, but if it works it works. 

If that's the primary change you're going to make to state stuff, then it should be pretty straightforward to integrate the state changes made here. 
When you make that PR I'll take a look and see if there's an easy opportunity for simplification/optimization.

AlexWKinley:
From my vantage point, I'm good with this being merged whenever. But I'm also happy to leave this open if you'd prefer to merge it after other work, or if we want to wait for Jose to take a look.

Thanks for all the work you've been doing on MoorDyn @RyanDavies19.

RyanDavies19:
This is good to merge on my end!

source/State.hpp
RyanDavies19: My first thought was to just change the vector sizes in the line state from vec3 for 5 (because these are technically line state variables) but everything in the state files is hard coded for vec3 or vec6, and it doesn't get around the fact that we only need to integrate these states once (unlike acceleration -> velocity -> position). I still need to develop tests for that work, and I am waiting on a preprint to link to explain the theory.

AlexWKinley:
@sanguinariojoe 

I agree it could be nice to better integrate with operator overloading. With how things currently work `butcher_row<1>(r[0], r[0], { dt }, { &rd[1] });` would still be more performant than `r[0] += rd[1] * dt`. Since rd[1] * dt would itself allocate a new state to store the scaled derivatives in. 

My future ideal would be for MoorDynState and MoorDynDState to subclass or internally wrap `Eigen::ArrayXd`. Because Eigen uses expression templates, things like `r[0] = r[0] + (rd[0] + rd[3]) * (dt / 6.0) + (rd[1] + rd[2]) * (dt / 3.0);` would avoid allocating and basically compile to the equivalent loop expression. But there's some complexity in terms of then being able to get individual object states, and the internal state code would get more complicated overall. So I've held off on attempting that, especially since there are other changes/additions to states in the pipeline from Ryan's work.

That being said, it's not hard to implement the `operator+=`, and if it would be useful I'm happy to add it. I just don't think any code currently uses it, since it's not currently defined.

sanguinariojoe:
Up to you Alex. As I said, this can be merged as it is.

You seem to have a plan, so I am totally ok with your design decisions.

I personally tend to avoid allocations abusing on += and *=, and sometimes
preallocating memory for + and *, overloading = as well to take advantage.
But I did not dive as much as you on the Eigen way, so I trust on you. On
top of that, you are usually the winning horse

RyanDavies19:
Let's merge this as in then for now if it works for you @AlexWKinley. One last thing before that, can we change `conn` to `point` in the `butcher_row` function (see review comment above)?

AlexWKinley:
Oops, I somehow missed that review, my apologies. Should be fixed now. I'm good for this to be merged.

Issue #278, opened 2024-11-28
Divergent branches


sanguinariojoe:
Hey @RyanDavies19 and @AlexWKinley !

For some reason (probably we rebased on master instead of merging) the dev branch has diverged from the master one. I rebased and the problem was automagically gone, so it is not really a big deal. However, if I do that I will break all your on-going work (including the active PRs).

Thus, we can proceed in two ways (I really think the first way is the good one):

1. Launch a new release forcily pushing the current dev on master
2. Rebase dev on master

The merit of way 1 is that everything will work just fine for all us. The problem is that we are forcily pushing on master, so some remote forks  on the wild might have problems to create PRs. I think the risk is rather low because we are actually the only active devs of this project. If someone complains on the close future I will take full responsibility on fixing the issue.

RyanDavies19:
I'm fine with force pushing to master. I am not aware of anyone doing and development, and if they are I would think they would be working off the dev branch.

sanguinariojoe:
OK, I force pushed. Patch release is on its way!

PR #279, opened 2024-11-28
Some minor changes to prepare for the 2.3.8 release

sanguinariojoe:
Just a couple of tweaks to launch the 2.3.8 release. Tomorrow I am still on the office. If we do not launch the missiles @RyanDavies19, we will have to wait until 12-dic (not really a drama anyway)


RyanDavies19:
@sanguinariojoe this all looks good, could we change the filename to polyester.py to keep things consistent? 

Also just to confirm, this python test is the same as the cpp test right? When I merge in the viscoelastic capability should the tests be put in both of these or is just the cpp one sufficient?

sanguinariojoe:
> @sanguinariojoe this all looks good, could we change the filename to polyester.py to keep things consistent?

I am afraid not. PyTest has the `test_*` naming policy (I suppose it can be done, but some hacking will be required that is probably backfiring on the future)

> Also just to confirm, this python test is the same as the cpp test right? When I merge in the viscoelastic capability should the tests be put in both of these or is just the cpp one sufficient?

cpp is surely enough

RyanDavies19:
@sanguinariojoe was this intended to trigger a release? Trying to work through some bugs with the 2.3.7 windows DLL causing GH actions to crash on the WECSim tests. I notices though we don't have any PR documenting the 2.3.8 release but the binaries are available.

sanguinariojoe:
Nope. It is because master and dev diverged, so I did the release forcily pushing on master. See https://github.com/FloatingArrayDesign/MoorDyn/issues/278

PR #280, opened 2024-12-05
fix: added dummy AllOutput to old API

RyanDavies19:
This adds a dummy AllOutput function to the old API wrapper that prints a message stating "In version 2, AllOutput is automatically called by MoorDynStep". This reduces warning messages when using the old API. This was seen when updating the WECSim libraries in https://github.com/WEC-Sim/MoorDyn/pull/7. 

@sanguinariojoe, let me know if there is a better way to print the message here. I didn't see anything in the C header files that allowed for the use of LOGMSG.

### DIFF
--- a/source/MoorDyn.cpp
+++ b/source/MoorDyn.cpp
@@ -287,5 +287,5 @@ AllOutput(double t, double dt)
 {
 	if (!md_singleton)
 		return;
-	std::cout << "In version 2, AllOutput is automatically called by MoorDynStep" << std::endl;
+	std::cout << "In version 2, AllOutput is automatically called by MoorDynInit and MoorDynStep" << std::endl;
 }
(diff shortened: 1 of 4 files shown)


RyanDavies19:
Apologies for the spam here with testing, windows compiling is my nemesis. 

In order to keep compilers happy, I renamed the AllOutput function to WriteOutputs. This matches what is in the MD-F code, and IMO is more clear about what is happening.

sanguinariojoe:
> @sanguinariojoe, let me know if there is a better way to print the message here. I didn't see anything in the C header files that allowed for the use of LOGMSG.

You can use [MoorDyn_Log](https://github.com/FloatingArrayDesign/MoorDyn/blob/master/source/MoorDyn2.h#L139)

RyanDavies19:
@sanguinariojoe just made that change, this should be good to go now

PR #284, opened 2025-01-02
Stiffness file reading correction

### LINKED ISSUE #283: Uncorrect reading of stiffness file
Using the C API to couple Moordyn and OpenFOAM I've found that the file defining non-linear stiffnes is not correctly read.
At the following [line](https://github.com/FloatingArrayDesign/MoorDyn/blob/4828499f8ade98521011f563c99fbd71c0429647/source/MoorDyn2.hpp#L759) seems that the same value is used for `x` and `y` values. If I'm not wrong, It should be set as

```
y.push_back(atof(entries[1].c_str()));
```

guillencampa:
Closes #283


sanguinariojoe:
TY!

PR #288, opened 2025-01-13
First Rust test!

sanguinariojoe:
It is quite ugly, but I do not know if it is worthy to create a thin wrapper to make it easier to use... Probably not...


RyanDavies19:
@sanguinariojoe this looks good to me (as someone who hasn't worked with Rust before). Thanks for putting it together. I think for the time being just checking things are basically working is fine, I don't know of any users of the wrapper at the moment so I don't think we need to dedicate too much effort to making it pretty looking.

PR #295, opened 2025-03-12
Vortex Induced Vibrations and Viscoelastic model for MoorDyn-C

RyanDavies19:
This is the flexstate compatible version of #290. Thanks for all the help @sanguinariojoe!

### New Features

This work contains two new features. Theory papers will be linked to this PR when publicly available.

1. The first is a vortex-induced-vibration model based on the work of [Thorsen et al.](https://www.sciencedirect.com/science/article/pii/S0951833916302118). We have a paper submitted for this and when a preprint is released I will link it here. The semi-empirical model gives an oscillating lift force on each node. As part of this implementation, there is a synchronization model that looks at changes between the lift force phase and line motion phase. This model requires knowledge of the lines current acceleration (which we approximate by the previously calculated value) and an additional state variable that is integrated every time step. We validated this approach against the results in the original theory paper using rk4. We did not test this approach with implicit integration schemes, though my hope with the acceleration approximation was that it would work in that case. I will add a warning that requires users to use explicit schemes.
2. The second is a pair of viscoelastic models that have been implemented in MoorDyn-F to simulate non-linear rope behavior. They are based on the idea of dividing a segment into a spring-dashpot in parallel in series with a second spring-dashpot in parallel (the [Kelvin representation of the standard linear solid model](https://en.wikipedia.org/wiki/Standard_linear_solid_model) in the case where dynamic damping = 0). A constant dynamic stiffness approach is explained in [this paper](https://doi.org/10.1115/IOWTC2023-119524). The mean-load dependent dynamic stiffness model builds off this work by changing the dynamic stiffness as a function of mean load. Test results for this are shown in the MoorDyn-F PR (OpenFAST/openfast#2459). Both of these require an additional state to handle the length of one of the springs in the subdivided segment. Long term, it would be great to allow for n-number of spring damper parallel components attached in series, which would require n more states per node. This would give us access to the true generalized Kelvin model. (A similar approach could be done with the generalized Maxwell viscoelastic model, also requiring n number of additional states per node). The mean-load dependent viscoelastic model internalizes the process of calculating the dynamic stiffness as a function of mean load. #194 introduced an API entry to enable this by using an external caller (as demonstrated in the polyester test), but that requires knowledge of the current tension to find the current stiffness and then adjusting the unstretched lengths of the segments. This approach allows the user to just provide the constant and load dependent terms of the viscoelastic equation as coefficients while MoorDyn internally handles the rest (without changing the unstretched length). 

The documentation updates reflect the input file changes that enable these new features. Other minor changes are:
1. \# is now a comment character in the input file and all text on a line after # is ignored. This matches MD-F
2. Added space before messages output by WaterKin to make the terminal output more clear
4. The VIV force and the line curvature at each node are additional line output flags

### State Approach

See #291. A note on the size of the line state instance: This implementation means that when simulating viscoelastic only, the line state instance is Nx7, when simulating VIV only N-1x7 and when simulating both VIV and viscoelastic Nx8. This does introduce unused spaces in the matrix, notably with the viscoelastic model. Indexing of the matrix is as follows (from comments in `Line::setState`): 

```
	// ----- State Instance Structure for Line -----
	// Each row corresponds to an internal node. When the viscoelastic case is 
	// activated the last column corresponds to the viscoelastic state for each segment. 
	// This means indexing the rows for the viscoelastic state is shifted by one.
	//
	// Normal case:
	//  - N-1 rows
	//	- row[i] = [rix, riy, riz, rdix, rdiy, rdiz]
	// VIV case:
	//  - N-1 rows
	//	- row[i] = [rix, riy, riz, rdix, rdiy, rdiz, phii]
	// Viscoelastic case:
	//  - N rows
	//	- row[i] = [rix, riy, riz, rdix, rdiy, rdiz, ldot_1i]
	//  - note there will be 6 unused values in the last row
	// VIV and viscoelastic case:
	//  - N rows
	//	- row[i] = [rix, riy, riz, rdix, rdiy, rdiz, phii, ldot_1i]
	//  - note there will be 7 unused values in the last row
```

This addresses the conversation started in #261 and https://github.com/FloatingArrayDesign/MoorDyn/pull/269#discussion_r1857480090. 

### Viscoelastic Verification

To verify the viscoelastic model, we replicated the results seen in OpenFAST/openfast#2459. This test introduced sinusoidal oscillations at different displacements to test the dynamic stiffness response. The figures below show the strain as a function of time and the stress strain curve from this test. The viscoelastic model impact on the dynamic stiffness is clear, at larger strains the line has a higher dynamic stiffness. This test has been added to the polyester tests.
mdc_strain
mdc_stress_strain

### TODO:
- [x] Resolve the failing tests
- [X] Develop test for VIV
- [x] Warning for users when using explicit schemes
- [x] Develop test for viscoelastic models 
- [X] Clean up and adapt to new state approach 
- [X] Clean up commit history to remove duplicates 
- [X] Ensure performance is retained with new state approach 
- [x] Handle defaults for viscoelastic and VIV states when loading in exported states (from MoorPy, Map++, or MoorDyn)


RyanDavies19:
Some more notes: I disabled all output time printing for the tests so that the logs are much easier to read. Also I adjusted spacing in messages printed by wavekin.

RyanDavies19:
@sanguinariojoe, this is ready for review when you get a chance. Also, any ideas of why tests are failing on Windows and Ubuntu but passing on MacOS? I am assuming some compiler differences, could it be an uninitialized variable or something similar?

sanguinariojoe:
> @sanguinariojoe, this is ready for review when you get a chance. Also, any ideas of why tests are failing on Windows and Ubuntu but passing on MacOS? I am assuming some compiler differences, could it be an uninitialized variable or something similar?

Yep, you have memory errors. I am checking them

source/Line.cpp
```diff
@@ -666,30 +726,70 @@ Line::setPin(std::vector<real> p)
...
+			else phi[i+1] = state.row(i).tail<1>()[0]; // else last element in the row
+		} 
+		
 	}
 }
 
 void
 Line::setEndKinematics(vec pos, vec vel, EndPoints end_point)
 {
+	if (t < 5 && int(t / dtm0) % 20 == 0) {
```
sanguinariojoe, line 787: This debugging chunk is using uninitialized values. Not much a problem though, but it is adding noise to the mem errors tracking

tests/state_tests.cpp
```diff
@@ -39,22 +39,24 @@ CreateLine(md::Log* log, EnvCondRef env, unsigned int n,
 	props.d = 0.1;
```
sanguinariojoe, line 39: `props.ElasticMod = 1;` is missing

source/Rod.cpp
sanguinariojoe, line 1: There seem to be a problem with fixed rods attached to bodies (see for instance the bodies test). The problem seems to be the sequence:

`moordyn::Body::initializeUnfreeBody()`->`moordyn::Rod::initialize()`->`moordyn::Rod::setDependentStates()`

where `moordyn::Rod::r7` is never initialized

sanguinariojoe:
@RyanDavies19 I marked you all the memory errors I have found. After fixing them, all the tests are completely clean of mem errors.

Regarding the time_schemes test, I would say it is just machine epsilon level errors. I have been prompting the states of the offending time integrator every single iteration, and everything is identical up to t = 0.0377 (29 steps).
The problem is that we are so close to the stability limit on that test... Just changing https://github.com/RyanDavies19/MoorDynC_ryan/blob/viv_flexstate/tests/time_schemes.cpp#L82 by

`                                      "1.2E-3",   // BEuler15`

fixes the problem on Linux.

source/Line.cpp
RyanDavies19: I just got rid of it, I forgot it was in there

RyanDavies19:
By some form of dark magic I don't understand, https://github.com/FloatingArrayDesign/MoorDyn/pull/295/commits/a5aacc3c087a2f097d2d0e14ceaea8716d24f074 seemed to fix the failing windows test. 

This is ready to merge after you review @sanguinariojoe.

tests/state_tests.cpp
RyanDavies19: Fixed

source/Line.hpp
```diff
@@ -321,14 +378,16 @@ class DECLDIR Line final : public Instance, public NatFreqCFL
...
 	 */
 	void setup(int number,
 	           LineProps* props,
 	           real l,
 	           unsigned int n,
 	           EnvCondRef env_in,
 	           shared_ptr<ofstream> outfile,
-	           string channels);
+	           string channels,
+			   real dtM0);
```
sanguinariojoe, line 390: Please, fix this indentation

source/Line.hpp
```diff
@@ -342,23 +401,51 @@ class DECLDIR Line final : public Instance, public NatFreqCFL
 		seafloor = seafloor_in;
 	}
 
-	/** @brief Compute the stationary Initial Condition (IC)
-	 * @param r The output state variable
-	 * @return The states, i.e. the positions of the internal nodes
-	 * (first) and the velocities of the internal nodes (second)
+	/** @brief Initializse the line object
```
sanguinariojoe, line 404: Initializse -> Initialize

source/Line.hpp
```diff
@@ -874,7 +965,10 @@ class DECLDIR Line final : public Instance, public NatFreqCFL
 	vec getEndSegmentMoment(EndPoints end_point, EndPoints rod_end_point) const;
 
 	/** @brief Calculate forces and get the derivative of the line's states
-	 * @param drdt Output state derivatives
+	* @param drdt Output state derivatives
```
sanguinariojoe, line 968: Indentation

source/Line.hpp
```diff
@@ -874,7 +965,10 @@ class DECLDIR Line final : public Instance, public NatFreqCFL
 	vec getEndSegmentMoment(EndPoints end_point, EndPoints rod_end_point) const;
 
 	/** @brief Calculate forces and get the derivative of the line's states
-	 * @param drdt Output state derivatives
+	* @param drdt Output state derivatives 
+	* @param vel Where to store the velocities of the internal nodes
```
sanguinariojoe, line 969: There are no vel, acc or misc parameters

source/Line.cpp
```diff
@@ -128,7 +129,8 @@ Line::setup(int number_in,
             unsigned int NumSegs,
             EnvCondRef env_in,
             shared_ptr<ofstream> outfile_pointer,
-            string channels_in)
+            string channels_in,
+			real dtM0)
```
sanguinariojoe, line 133: Indentation

sanguinariojoe:
Everything looks fine on my side. I was just checking that the IC is not messing up, but it seems that you have it well controlled with the IC_gen and t_old vars.

So green light from my side, you can merge

source/Line.cpp
RyanDavies19: Fixed with https://github.com/FloatingArrayDesign/MoorDyn/pull/295/commits/183701dabdbd54fbd3566acd841301a8022bd19b

source/Line.hpp
RyanDavies19: Fixed with https://github.com/FloatingArrayDesign/MoorDyn/pull/295/commits/183701dabdbd54fbd3566acd841301a8022bd19b

source/Line.hpp
RyanDavies19: Fixed with https://github.com/FloatingArrayDesign/MoorDyn/pull/295/commits/183701dabdbd54fbd3566acd841301a8022bd19b

RyanDavies19:
> Everything looks fine on my side. I was just checking that the IC is not messing up, but it seems that you have it well controlled with the IC_gen and t_old vars.
> 
> So green light from my side, you can merge

Great thanks, merging! Yeah without the IC_gen flag and t_old it would never initialize in steady state because of the quasi-chaotic nature of the model (especially when currents were involved). It does mean that the first few seconds of a simulation with VIV will show the model kicking in and the system approaching a lock-in state.

RyanDavies19:
Linking the theory and validation paper: https://doi.org/10.1016/j.jfluidstructs.2026.104651

Issue #327, opened 2025-09-17
MoorDyn DLLs crashing MATLAB on GitHub actions


jtgrasb:
For WEC-Sim, we are using MoorDyn dynamic link libraries to call MoorDyn and return the results to the WEC-Sim simulation. The dynamic link libraries are built using [cmake via GitHub actions](https://github.com/jtgrasb/MoorDyn/blob/main/.github/workflows/compileLibraries.yml) and build with no issues. We have this running well locally, but the MoorDyn DLLs seem to cause GitHub actions to fail by crashing MATLAB. See [the latest test on our dev branch of WEC-Sim Applications](https://github.com/WEC-Sim/WEC-Sim_Applications/actions/runs/17800892973/job/50600275103). If I rerun this a few times, it will pass, but it is still frustrating to have to rerun it all the time.

I *think* the MATLAB crash may be caused by a window that MoorDyn tries to open. In MATLAB locally, the window shows up like below, but I think it may be trying to display some sort of graphic? While running the simulation locally does not crash MATLAB, closing this window does crash MATLAB. Do you know what may be causing this window to be opened? And if so, is there a way to disable it?

<img width="400" height="253" alt="Image" src="https://github.com/user-attachments/assets/4337745f-89cd-4c07-8c45-d1359643ff4a" />

--Thanks,
Jeff

jtgrasb:
Got it, that makes sense. Thank you! I believe we stuck with MoorDyn v1 API to avoid extra files. MoorDyn2.h calls Waves,h, Seafloor.h, etc. but we don't use these at all in WEC-Sim. While its not a big deal to add the other .h files, the only functions that we use are MoorDyn_Init, MoorDyn_Create, MoorDyn_Step, and MoorDyn_Close so it'd be nice to not include extra files unnecessarily. I edited MoorDyn2.h to remove all other functions except these 4 and it seems to work fine for our purposes. Are there any issues you see with this approach?

jtgrasb:
@RyanDavies19 Got it, thanks! I think updating to the V2 version makes sense. Since were still using `MoorDynAPI.h` and the dll is being updated for every version, there shouldn't be any significant changes in the future unless those 4 function names/inputs are changed, right?

sanguinariojoe:
If we do an API change we are also changing the version. In particular:

- Added feature = minor version change
- Breaking change = major version change

RyanDavies19:
[...]

PR #333, opened 2025-10-22
Add function to disable console window

jtgrasb:
This PR adds a function that can be called to disable the console window when using the v1 API. This will allow WEC-Sim to disable the console window before initializing MoorDyn, hopefully resolving issues with MATLAB crashing. See #327.


sanguinariojoe:
The pipeline is failing... But it has been failing since 28th August. I cannot see any commit that might triggered the issue. We should investigate that @RyanDavies19 

Anyway, @jtgrasb increase the version [here](https://github.com/jtgrasb/MoorDyn_Official/blob/disable_console/CMakeLists.txt#L3) to 2.5.0 please. After that you can proceed to squash and merge guys (remember to put a sensible commit message according to [the conventional commits standard](https://www.conventionalcommits.org/en/v1.0.0/))

jtgrasb:
@sanguinariojoe @RyanDavies19 Thanks for the help. I think this PR should be good to merge.

sanguinariojoe:
OK, Apparently the CMake fails are caused by GitHUB actions shifting macos-latest to arm64 arch (see [here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)). Why is that a problem for math? no idea...

Anyway, Both the Python wrapper and the Matlab wrapper shall be running just fine:

https://github.com/core-marine-dev/MoorDyn/actions/runs/19419494577
https://github.com/core-marine-dev/MoorDyn/actions/runs/19419494575/job/55554020202

So there is something introduced here which is breaking the wrappers.

DO NOT MERGE UNTIL FIXED PLEASE.

I will first try to fix the problems on the CI/CD and then I will focus on this (if not fixed yet)

sanguinariojoe:
@RyanDavies19 , you should squash and merge, not merge directly. And you should set a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/).

I am adding an useless commit to document the changes

RyanDavies19:
@sanguinariojoe thanks for the reminder on conventional commits. @jtgrasb do you need a release for WECSim?

jtgrasb:
@RyanDavies19 @sanguinariojoe I'm testing this out now. I was thinking I could just update our MoorDyn.h and MoorDynAPI.h, then set disableConsole equal to 1 in the lines.txt file, but that is not working. Is there another way to set the disableConsole variable?

`WRN D:\a\MoorDyn\MoorDyn\source\MoorDyn2.cpp:2443 moordyn::MoorDyn::readOptionsLine(): Warning: Unrecognized option 'disableconsole'`

Issue #355, opened 2026-01-10
v2.5.0 release


RyanDavies19:
Hi @sanguinariojoe, 

What do we need to do to release v2.5.0? In #333 you said there was some work that needed to be done on the GH actions side of things. 

Asking because the version increase is needed for the bug fixes in the WEC+Sim to MoorDyn coupling @jtgrasb has been working on.

sanguinariojoe:
Well, this should not be an issue anymore! The new release is ready!

https://github.com/FloatingArrayDesign/MoorDyn/releases/tag/v2.6.0
https://pypi.org/project/moordyn/2.6.0/

PR #359, opened 2026-02-12
feat(core): No more huge dependency VTK

sanguinariojoe:
This is called to fix https://github.com/FloatingArrayDesign/MoorDyn/issues/355

As side effects:

 - ~Now the files are not .vtp but .vtu. That matters because Paraview does not automatically switchs wireframe visualization, so if you load them you will see nothing (that happened actually on some version of Paraview even with the .vtp)~
 - We have lost the `MoorDyn_UseBodyVTK()`, which I suppose is not that a big deal

### DIFF
--- a/source/Body.hpp
+++ b/source/Body.hpp
@@ -438,7 +438,7 @@ class DECLDIR Body final
 	 */
 	uint64_t* Deserialize(const uint64_t* data);
 
-	/** @brief Save the body on a VTK (.vtu) file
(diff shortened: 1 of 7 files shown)


sanguinariojoe:
Well, it is true that we lost `MoorDyn_UseBodyVTK()`, but we have recovered the VTPs!

I am merging and preparing everything for the release!

RyanDavies19:
Fixed by #361, thanks for doing that

PR #360, opened 2026-02-17
v2.6.0

sanguinariojoe:
v2.6.0 release!


RyanDavies19:
Thanks @sanguinariojoe. Just documenting in the comments here that this release was to change from `VTK` to `lean-vtk`, which resolves issues that were causing python wheels to crash in GH actions (see #355).

PR #369, opened 2026-03-19
Implementation of the Syrope model for polyester ropes in MoorDyn

zhilongwei:
**Feature or improvement description**

This pull request implements the Syrope model for polyester mooring ropes in MoorDyn-C.

The implementation captures:
- nonlinear, path-dependent, and loading-rate-depedent tension–strain behaviour of polyester ropes,
- dependence of dynamic stiffness on mean tension (Wave-frequecy and low-frequency regimes),
- working-curve-based quasi-static response.

**Impacted areas of the software**

- Mooring line constitutive model (new rope type: Syrope-based polyester rope)
- Time-domain solver (integration of visco-elasto-plastic response)
- Input file handling (additional parameters for Syrope model)
- Documentation

**Additional supporting information**

- A new regression test case, `syrope`, has been introduced. A standalone version of the test is also available at: https://github.com/zhilongwei/moordyn-syrope-tests.git, along with example input files for the Syrope model.
- Documentation has been updated to describe the model formulation and usage.
- A detailed reference for the numerical implementation is provided in:
  [ESOMOOR D5.1: Extended MoorDyn solver and validation report](https://doi.org/10.11583/DTU.31408806).

**Generative AI usage**

- Assisted by: GitHub Copilot <support@github.com>

**Test results**

- All existing and new tests pass.

**Notes for reviewers**

- A new `elastic_model` type, `ELASTIC_SYROPE = 4`, has been added alongside the existing three models. It is triggered exclusively when the Syrope model is specified through the `EA` field in the input file.
- No modifications have been made to existing line types.
- Full backward compatibility is maintained.


tests/syrope.cpp
```diff
@@ -0,0 +1,628 @@
...
+			else
+				tension_wc = Tmax * (std::exp(k2 * xi) - 1.0) / den;
+			break;
+		}
+		default:
+			throw std::invalid_argument(
+			    "find_mean_tension: unknown WorkingCurveForm");
+	}
+
+	if (std::abs(tension_wc - Tmax) <= 1e-12 * std::max(1.0, Tmax)) {
```
sanguinariojoe: `std::max(1.0, Tmax)` -> `(std::max)(1.0, Tmax)` for Windows sake

tests/syrope.cpp
```diff
@@ -0,0 +1,628 @@
...
+ */
+
+#include "MoorDyn2.h"
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <Eigen/Dense>
+
+#include <algorithm>
+#include <cmath>
```
sanguinariojoe, line 17: `#define _USE_MATH_DEFINES` before including cmath, otherwise Windows is not defining the constants.

To be sure that cmath is not included elsewhere before, voiding the definition, I suggest defining it at line 8

tests/syrope.cpp
```diff
@@ -0,0 +1,628 @@
...
+	out.times = times;
+	out.strain = strain;
+	out.tension_output = load_seg_te_column(line1_output_from_input(c.input_file));
+	return out;
+}
+
+double
+relative_l2(const Eigen::VectorXd& ref, const Eigen::VectorXd& val)
+{
+	const Eigen::Index n = std::min(ref.size(), val.size());
```
sanguinariojoe: `std::min(ref.size(), val.size())` -> `(std::min)(ref.size(), val.size())` for Windows sake (again).

Copilot helped to craft this, and it does not know the basic Windows pitfalls... Woha Microsoft! what a nice product you are developing!

tests/syrope.cpp
```diff
@@ -0,0 +1,628 @@
...
+	};
+
+	const bool superimpose_fast = true;
+
+	for (const auto& c : cases) {
+		DYNAMIC_SECTION("Case: " << c.name)
+		{
+			const auto sim = run_case(c, superimpose_fast);
+			const Eigen::Index n =
+			    std::min(sim.tension_output.size(), sim.strain.size());
```
sanguinariojoe: IDEM

tests/syrope.cpp
```diff
@@ -0,0 +1,628 @@
...
+			REQUIRE(n >= 2);
+
+			Eigen::VectorXd tmax_mean(n);
+			tmax_mean[0] = kTmax0;
+			for (Eigen::Index i = 1; i < n; ++i) {
+				tmax_mean[i] = std::max(tmax_mean[i - 1], sim.tension_output[i]);
```
sanguinariojoe: IDEM

tests/Mooring/syrope/quadratic_wc/line.txt
```diff
@@ -0,0 +1,35 @@
+One single segment.
+------------------------- LINE TYPES --------------------------------------------------
+LineType  Diam    MassDenInAir    EA       BA/-zeta    EI    Can   Cat  Cdn   Cdt
+(-)       (m)       (kg/m)        (-)      (Pa-s/-)  (n-m^2) (-)   (-)  (-)   (-)
+poly      0.203     40      SYROPE:../owc.dat|1.53e8|23.12    5.0e10|1.0e5      0.0   0.0   0.0   0.0  0.0
```
sanguinariojoe: Not really a big fan on how syrope config is set, specially those new full sections. I will let anyway @RyanDavies19 @mattEhall and @shousner to decide

zhilongwei: Thanks for the feedback. The additional sections were introduced to keep all Syrope model parameters within the existing input file, avoiding the need for a separate configuration file. This may not be the most consistent approach with the current input structure. An alternative would be to move the Syrope-related parameters into a separate file if that better aligns with the overall design.

sanguinariojoe:
BTW @zhilongwei , please direct the PR to the dev branch (and rebase with it) please

zhilongwei:
It seems that the new test `syrope` takes a long time to finish, which also failed locally with `ctest -C RelWithDebInfo -T memcheck -j 4 -R syrope`. I will try to reduce the simulation duration and see. I am not sure why the MoorDynF regression tests are failing, as they do not appear to be related to the changes in MoorDynC.

sanguinariojoe:
Do kot worry about the test failing. Ryan will fix that eventually

sanguinariojoe:
@zhilongwei memcheck can be quite time consuming. If your test does not fit, it is not really a big deal. Just add it to the list of excluded tests here:

https://github.com/FloatingArrayDesign/MoorDyn/blob/master/.github/workflows/memcheck.yml#L39

I will run it anyway with memcheck to be sure and let you know

sanguinariojoe:
Yep, a lot of beef. The stdout of the test:

```
18: Working Directory: /home/pepe/CoreMarine/MoorDyn/MoorDyn.gcc/tests
18: Test timeout computed to be: 1500
18: Randomness seeded to: 2632179782
18: 
18:  Running MoorDyn (v2.6.2)
18:          MoorDyn v2 has significant ongoing input file changes from v1.
18:    Copyright: (C) 2024 National Renewable Energy Laboratory, (C) 2014-2019 Matt Hall
18:    This program is released under the  BSD 3-Clause license.
18: The filename is Mooring/syrope/linear_wc/line.txt
18: MoorDyn v2 log file with output level MSG at 'Mooring/syrope/linear_wc/line.log'
18: Loading a curve from 'Mooring/syrope/linear_wc/../owc.dat'...
18: 29 lines of curve successfully loaded
18: Generated entities:
18:     nLineTypes  = 1
18:     nRodTypes   = 0
18:     nPoints     = 2
18:     nBodies     = 0
18:     nRods       = 0
18:     nLines      = 1
18:     nFails      = 0
18:     nFreeBodies = 0
18:     nFreeRods   = 0
18:     nFreePoints  = 0
18:     nCpldBodies = 0
18:     nCpldRods   = 0
18:     nCpldPoints = 1
18: Time integrator = 2nd order Runge-Kutta
18: Water kinematics for IC gen:
18:    No Waves or Currents, or set externally
18: Creating mooring system...
18: Initializing coupled Point 2 at 1.01478, 0, -5
18: dtM = 0.01 s (CFL = 2.15154)
18: Finalizing ICs using static solve
18:   - Line1:
18:     ID      : 1
18:     UnstrLen: 1
18:     N       : 1
18:     d       : 0.203
18:     rho     : 1235.88
18:     EAMod   : 4
18:     EA      : 0
18:     BA      : 5e+10
18:     EI      : 0
18:     Can     : 0
18:     Cat     : 0
18:     Cdn     : 0
18:     Cdt     : 0
18:     Cl      : 0
18:     dF      : 0
18:     cF      : 0
18:     ww_l    : 66.9571
18: WRN /home/pepe/CoreMarine/MoorDyn/MoorDyn/source/Line.cpp:676 initialize(): Catenary initial profile failed for Line 1, initalizing as linear 
18: Initialized Line 1
18: Remaining error after 0.01 s = 0 m/s2
18: Best score at 0.01 s = 0 m/s2
18: Water kinematics for runtime:
18:    No Waves or Currents, or set externally
18: /usr/include/c++/15.2.1/bits/stl_vector.h:1282: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::operator[](size_type) const [with _Tp = Eigen::Matrix<double, 3, 1>; _Alloc = std::allocator<Eigen::Matrix<double, 3, 1> >; const_reference = const Eigen::Matrix<double, 3, 1>&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
18: 
18: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
18: syrope is a Catch2 v3.3.2 host application.
18: Run with -? for options
18: 
18: -------------------------------------------------------------------------------
18: Syrope tests
18:   Case: Linear
18: -------------------------------------------------------------------------------
18: /home/pepe/CoreMarine/MoorDyn/MoorDyn/tests/syrope.cpp:601
18: ...............................................................................
18: 
18: /home/pepe/CoreMarine/MoorDyn/MoorDyn/tests/syrope.cpp:601: FAILED:
18:   {Unknown expression after the reported line}
18: due to a fatal error condition:
18:   SIGABRT - Abort (abnormal termination) signal
18: 
18: ===============================================================================
18: test cases: 1 | 1 failed
18: assertions: 5 | 4 passed | 1 failed
18: 
1/1 MemCheck #18: syrope ...........................Subprocess aborted***Exception: 1398.72 sec
18: process test output now: syrope syrope
PostProcessTest memcheck results for : syrope

0% tests passed, 1 tests failed out of 1

Total Test time (real) = 1398.72 sec

The following tests FAILED:
         18 - syrope (Subprocess aborted)
```

The memory errors:

[MemoryChecker.18.log](https://github.com/user-attachments/files/26207208/MemoryChecker.18.log)

I see a lot of problems reading the config and setting up the system. For the time being, please rebase with `dev` branch, so at least the VTK issues will be knocked out. When the time allows I will check your changes and give some more feedback about all these errors

zhilongwei:
Hi @sanguinariojoe, thank you very much. I have rebased with the `dev` branch.

sanguinariojoe:
OK, I found the main problem @zhilongwei and fixed it on #371 . I am running the memcheck again. Meanwhile please rebase again to `dev` branch.

Then you can try to reduce the simulation time (even on the regular environment your new test is taking 60 seconds), or drop it from the memcheck tests.

zhilongwei:
I will try to reduce the simulation time. Currently, for each working curve formula (linear, quadratic, exponential), the test includes six phases of 3 hours each to mimic realistic loading scenarios. This may not be necessary.

zhilongwei:
I have reduced the memory-check runtime to about 4 minutes on my local machine. I hope this is acceptable. If not, I can exclude this test from the memcheck tests.

sanguinariojoe:
> I have reduced the memory-check runtime to about 4 minutes on my local machine. I hope this is acceptable. If not, I can exclude this test from the memcheck tests.

That should do the job. Anyway, we will see when launching the next release. If it is problematic I am droping it myself.

For me this is good to go. So let's wait for @RyanDavies19 @mattEhall and @shousner to give their feedback.

tests/Mooring/syrope/quadratic_wc/line.txt
RyanDavies19: @zhilongwei, I think it would be better if all the Syrope inputs were a separate input file, which would be consistent with the approach used for a non-linear lookup table. 

Then the main moordyn input file would read:
```
------------------------- LINE TYPES --------------------------------------------------
LineType  Diam   MassDenInAir        EA                   BA/-zeta      EI     Can   Cat  Cdn   Cdt
(-)       (m)     (kg/m)             (-)                  (Pa-s/-)    (n-m^2)  (-)   (-)  (-)   (-)
poly      0.203     40    SYROPE:input.dat|1.53e8|23.12  5.0e10|1.0e5  0.0    0.0   0.0   0.0  0.0
```

And then `input.dat` would read: 
```
../owc.dat    nonlinear stiffness lookup table path
-------------------------- SYROPE WORKING CURVES ------------------------------------------------- 
LineType   WCType    k1      k2
(-)        (-)       (-)     (-)
poly      QUADRATIC  0.25     1.0
-------------------------- SYROPE IC ------------------------------------------------- 
Line     Tmax     Tmean 
(-)      (N)       (N)
1        1.3886e6   4.339e5
```

Note `input.dat` is just a placeholder, you could call the file whatever you wanted to.

docs/inputs.rst
```diff
@@ -333,6 +333,65 @@ needs to be input as a value. Example inputs are below:
...
+
+Example:
+
+.. code-block:: none
+
+  TypeName   Diam    Mass/m     EA                               BA
+  (name)     (m)     (kg/m)     (N)                              (N-s)
+  poly       ...     ...        SYROPE:owc.txt|EA_Dc|EA_D_Lm     BA_s|BA_d
+
+The file ``owc.txt`` follows the same format as the tabulated non-linear stiffness files described
```
RyanDavies19: When updating the docs for a separate input file for the Syrope model, this explanation should be moved to a new section in [Additional MoorDyn Files](https://moordyn.readthedocs.io/en/latest/inputs.html#additional-moordyn-files). 

If you are able to also move the explanation of the non-linear stiffness lookup table input file (lines 310-320) to a new section in [Additional MoorDyn Files](https://moordyn.readthedocs.io/en/latest/inputs.html#additional-moordyn-files) that would be great (that has been a TODO for a while but we haven't gotten around to it). 

Ideally, [The V2 Input File](https://moordyn.readthedocs.io/en/latest/inputs.html#the-v2-input-file) section only contains explanations of the main MoorDyn input file. And then [Additional MoorDyn Files](https://moordyn.readthedocs.io/en/latest/inputs.html#additional-moordyn-files) is a catchall for the additional sub models' input files that are added into MoorDyn.

docs/references.rst
```diff
@@ -107,6 +107,10 @@ Viscoelastic approach for non-linear rope behavior:
   `Hall, Matthew, Brian Duong, and Ericka Lozon, “Streamlined Loads Analysis of Floating Wind Turbines With Fiber Rope Mooring Lines.” In ASME 2023 
   5th International Offshore Wind Technical Conference, V001T01A029. Exeter, UK: American Society of Mechanical Engineers, 2023. <https://doi.org/10.1115/IOWTC2023-119524>`_
 
+Syrope model for polyester ropes:
```
RyanDavies19: Can you move this to underneath `Modeling of Bi-stable Nonlinear Energy Sinks in MoorDyn`? The references here should be listed in chronological order.

To that end, I think `Implementation of bending stiffness modeling for power cables` needs to be moved above `Seabed friction and bathymetry approach used in v2`. Feel free to make this change if you want.

RyanDavies19:
Hi all,

First off, the report is a great documentation of the code changes and validation. It will definitely be a useful reference for any future work building off of this. Thanks for putting it together!

I left a few comments on input file formatting and documentation. I don't have time for a full code review, but I trust that @sanguinariojoe can find any major issues. After the docs changes are made this looks good to merge to me too!

@zhilongwei, based on your report I am assuming there is a concurrent PR being prepared for OpenFAST?

docs/inputs.rst
RyanDavies19: Change the short description of `EA` to the following:

`the line stiffness, product of elasticity modulus and cross-sectional area (N)` --> `the line stiffness, product of elasticity modulus and cross-sectional area (N), or nonlinear parameters (see descriptions below)`

RyanDavies19: Once you add the Syrope section to the Additional MoorDyn Files, can you add the following after the equation sentence (with a link to the Syrope section): `This method can be used standalone, and is also part of the Syrope model described in <link to syrope in the Additional MoorDyn Files section>.` 


Change: `is the slope of the load-stiffness curve. Both of these methods allow users to provide static` --> `is the slope of the load-stiffness curve. This method can be used standalone, and is also part of the Syrope model described in <link to syrope in the Additional MoorDyn Files section>. Both of these methods allow users to provide static`

zhilongwei:
> Hi all,
> 
> First off, the report is a great documentation of the code changes and validation. It will definitely be a useful reference for any future work building off of this. Thanks for putting it together!
> 
> I left a few comments on input file formatting and documentation. I don't have time for a full code review, but I trust that @sanguinariojoe can find any major issues. After the docs changes are made this looks good to merge to me too!
> 
> @zhilongwei, based on your report I am assuming there is a concurrent PR being prepared for OpenFAST?

Thank you, @RyanDavies19. Yes, we have a concurrent PR for MoorDyn-F in OpenFAST that is nearly ready. We still need to update the Syrope input format there to align with the changes here. Once everything is finalized here, we will submit a corresponding PR to OpenFAST with the same functionality.

tests/Mooring/syrope/quadratic_wc/line.txt
zhilongwei: I am now thinking that the initial Tmax and Tmean should not be treated as line-type properties since they are line-specific. For example, different lines that share the same Syrope properties (e.g., original working curve, working curves, etc.) may still have different initial values of Tmax and Tmean. It may therefore be more appropriate to separate these parameters from the Syrope input file (`input.dat`). One idea is to add two additional fields in the Lines section (after `LineOutputs`). These fields would only be required and read when the corresponding line uses a Syrope-based line type.

An example is shown below.
```
------------------------- LINE TYPES --------------------------------------------------
LineType  Diam   MassDenInAir        EA                   BA/-zeta      EI     Can   Cat  Cdn   Cdt
(-)       (m)     (kg/m)             (-)                  (Pa-s/-)    (n-m^2)  (-)   (-)  (-)   (-)
poly      0.203     40    SYROPE:input.dat|1.53e8|23.12  5.0e10|1.0e5  0.0    0.0   0.0   0.0  0.0
```

The `input.dat` reads
```
-------------------------- SYROPE ORIGINAL WORKING CURVE -----------------------------------------
../owc.dat    nonlinear stiffness lookup table path
-------------------------- SYROPE WORKING CURVES ------------------------------------------------- 
LineType   WCType    k1      k2
(-)        (-)       (-)     (-)
poly      QUADRATIC  0.25     1.0
--------------------------- need this line -------------------------------------------------
```

And the `Line` section reads
```
-------------------------- LINES -------------------------------------------------
Line     LineType NodeA     NodeB  UnstrLen   NumSegs     Flags/Outputs   Tmax   Tmean
(-)      (-)       (-)       (-)    (m)         (-)          (-)          (N)     (N)
1        poly       2         1   1.0          1             tsc        1.5e6   8.0e5
2        poly       4         3   1.0          1             tsc        2.0e6   5.0e5
3        chain      6         5   1.0          1             tsc
```

Let me know your thoughts on this.

RyanDavies19: I think that could work well. We already do a similar thing for VIV in the line properties section. I would say that if we do this, error handling should be added that logs when the syrope model is being used and throws a non-fatal warning if the Tmax and Tmin columns are detected without the syrope model activated. Similarly a fatal error should be thrown if Syrope is activated and these are not found. 

And just for clean looking input files, in the docs and test cases, you should add a `-` in the column for a non-syrope lines:

```
-------------------------- LINES -------------------------------------------------
Line     LineType NodeA     NodeB  UnstrLen   NumSegs     Flags/Outputs   Tmax   Tmean
(-)      (-)       (-)       (-)    (m)         (-)          (-)          (N)     (N)
1        poly       2         1   1.0          1             tsc        1.5e6   8.0e5
2        poly       4         3   1.0          1             tsc        2.0e6   5.0e5
3        chain      6         5   1.0          1             tsc          -       -
```

mattEhall:
Hi all,
Thanks for this great addition and the work on integrating it, and sorry I'm late to the discussion.

This is a fantastic new capability to have, and a good step in supporting more advanced options for elasticity. (I didn't come up with an approach for integrating a previous advanced elasticity model from @jnvn7.) The approach used here could be a good template to follow for if there other elasticity models added in the future as well.

I appreciate the discussion about the input files. I like the approach of linking to a separate model-specific input file for the reasons you discussed.

I feel hesitant about adding Tmax and Tmean as new core inputs in the LINES section because (1) they change the core input file format, (2) they aren't used in some cases and (3) they're more of a state/status than a design characteristic. I think having a separate SYROPE IC section for them in the syrope.dat file, in the format discussed earlier, is the better approach. E.g.:
```
-------------------------- SYROPE IC ------------------------------------------------- 
Line(s)     Tmax     Tmean 
(-)         (N)       (N)
1,2,3        1.3886e6   4.339e5
```
In this approach, the first column could also support a _list of lines_ (similar to the format we use for specifying line failures), for user convenience.

sanguinariojoe:
Guys, I have been thinking... From the following scheme on the report:

image

It seems reasonable to let `MoorDyn_Init()` guess the initial values, and then provide API entries so the user can modify those values. Something like `MoorDyn_SetLineSyrope()`

I really think that is the sensible way to proceed on MoorDynC, but it might be problematic on MoorDynF

zhilongwei:
> Hi all, Thanks for this great addition and the work on integrating it, and sorry I'm late to the discussion.
> 
> This is a fantastic new capability to have, and a good step in supporting more advanced options for elasticity. (I didn't come up with an approach for integrating a previous advanced elasticity model from @jnvn7.) The approach used here could be a good template to follow for if there other elasticity models added in the future as well.
> 
> I appreciate the discussion about the input files. I like the approach of linking to a separate model-specific input file for the reasons you discussed.
> 
> I feel hesitant about adding Tmax and Tmean as new core inputs in the LINES section because (1) they change the core input file format, (2) they aren't used in some cases and (3) they're more of a state/status than a design characteristic. I think having a separate SYROPE IC section for them in the syrope.dat file, in the format discussed earlier, is the better approach. E.g.:
> 
> ```
> -------------------------- SYROPE IC ------------------------------------------------- 
> Line(s)     Tmax     Tmean 
> (-)         (N)       (N)
> 1,2,3        1.3886e6   4.339e5
> ```
> 
> In this approach, the first column could also support a _list of lines_ (similar to the format we use for specifying line failures), for user convenience.

Thank you @mattEhall. This approach is feasible. However, it would require defining Tmax and Tmean for Syrope lines (in the LINE TYPES section) before the LINES section. As you pointed out, Tmax and Tmean are state variables rather than design parameters (or line properties), so they should ideally be defined separately from the Syrope properties. To preserve the core input file structure, it may still be acceptable to include them alongside the Syrope working curves, provided this is clearly documented and does not introduce confusion for users.

zhilongwei:
> Guys, I have been thinking... From the following scheme on the report:
> 
> image
> It seems reasonable to let `MoorDyn_Init()` guess the initial values, and then provide API entries so the user can modify those values. Something like `MoorDyn_SetLineSyrope()`
> 
> I really think that is the sensible way to proceed on MoorDynC, but it might be problematic on MoorDynF

Hi @sanguinariojoe, thank you for the suggestion. The method here (as in SIMA/RIFLEX), however, is a multi-stage method and requires a prior static analysis to find Tmean only. Tmax would still need to be provided. As MoorDyn is a time-domain solver, it is unable to guess the initial values.

sanguinariojoe:
> > Guys, I have been thinking... From the following scheme on the report:
> > image
> > It seems reasonable to let `MoorDyn_Init()` guess the initial values, and then provide API entries so the user can modify those values. Something like `MoorDyn_SetLineSyrope()`
> > I really think that is the sensible way to proceed on MoorDynC, but it might be problematic on MoorDynF
> 
> Hi @sanguinariojoe, thank you for the suggestion. The method here (as in SIMA/RIFLEX), however, is a multi-stage method and requires a prior static analysis to find Tmean only. Tmax would still need to be provided. As MoorDyn is a time-domain solver, it is unable to guess the initial values.

Well, it was worthy to ask!

RyanDavies19 (approved):
Thanks for moving things around in the docs. All this looks good to me! I will defer to @mattEhall for the `TMax0` and `TMean0` placement location

zhilongwei:
Hi all,

I am still hesitating about the suggestion to place Tmax and Tmean in the same file as Syrope working curves.

* Tmax and Tmean are state variables associated with individual lines, rather than intrinsic material or constitutive properties. Grouping them with the working curves may therefore be conceptually confusing.
* Defining Tmax and Tmean within the Syrope input file requires associating them with specific line IDs. However, the mapping between lines and line types is already defined in the LINES section. This leads to a duplicated definition of line associations across two locations, which can introduce ambiguity, increase the risk of inconsistency, and make the input structure harder to understand and maintain.

As an alternative, one possible approach is to introduce an optional SYROPE IC section in the main input file (similar to the FAILURE section), placed after the LINES section. For example:
```
-------------------------- SYROPE IC ------------------------------------------------- 
Line    Tmax     Tmean 
(-)      (N)       (N)
1,2,3   1.4e6    4.3e5
```
This section is only required when Syrope is used, and a warning will be raised if it is present but no Syrope line types are defined in the LINE TYPES section.

I would appreciate your thoughts on this approach.

Below is a realistic example that has been tested and run successfully:
```
--------------------- MoorDyn Input File ------------------------------------
Use polyester ropes with Syrope model
FALSE    Echo      - echo the input file data (flag)                                            
----------------------- LINE TYPES ------------------------------------------
TypeName  Diam    Mass/m    EA                              BA/-zeta     EI       Cd    Ca    CdAx  CaAx  Cl   dF   cF
(name)    (m)     (kg/m)    (N)                             (N-s/-)      (N-m^2)  (-)   (-)   (-)   (-)   (-)  (-)  (-)
chain     0.333   684.50    2.929E+09                       -1.0          0        2.4   1.00  0.0   0.0   0.0  0.0  0.0
poly      0.203    44.60    SYROPE:syrope.dat|4.16e8|23.13  1.0e11|1.0e6  0        1.6   1.00  0.0   0.0   0.0  0.0  0.0
-------------------------------- POINTS --------------------------------
ID   Attachment   X         Y         Z      Mass  Volume  CdA    Ca
(#)  (word/ID)   (m)       (m)       (m)    (kg)   (mˆ3)   (m^2)  (-)
1    Vessel    -54.875    -5.413   -14.000   0.0   0.0     0.0    0.0
2    Vessel    -54.875     5.413   -14.000   0.0   0.0     0.0    0.0
3    Vessel     22.750    50.230   -14.000   0.0   0.0     0.0    0.0
4    Vessel     32.125    44.817   -14.000   0.0   0.0     0.0    0.0
5    Vessel     32.125   -44.817   -14.000   0.0   0.0     0.0    0.0
6    Vessel     22.750   -50.230   -14.000   0.0   0.0     0.0    0.0
7    Free      -80.970     0.000   -27.730   0.0   0.0     0.0    0.0
8    Free       40.480    70.120   -27.730   0.0   0.0     0.0    0.0
9    Free       40.480   -70.120   -27.730   0.0   0.0     0.0    0.0
10   Free     -741.571     0.000  -206.297   0.0   0.0     0.0    0.0
11   Free      370.786   642.220  -206.297   0.0   0.0     0.0    0.0
12   Free      370.786  -642.220  -206.297   0.0   0.0     0.0    0.0
13   Fixed    -873.710     0.000  -243.300   0.0   0.0     0.0    0.0
14   Fixed     436.855   756.655  -243.300   0.0   0.0     0.0    0.0
15   Fixed     436.855  -756.655  -243.300   0.0   0.0     0.0    0.0
---------------------- LINES --------------------------------------                                           
ID   LineType  AttachA  AttachB  UnstrLen  NumSegs  LineOutputs
(#)   (name)    (ID)     (ID)      (m)       (-)      (-)
1    chain       7         1       30.000    20        -
2    chain       7         2       30.000    20        -
3    chain       8         3       30.000    20        -
4    chain       8         4       30.000    20        -
5    chain       9         5       30.000    20        -
6    chain       9         6       30.000    20        -
7     poly      10         7      681.240    80        -
8     poly      11         8      681.240    80        -
9     poly      12         9      681.240    80        -
10   chain      13        10      140.800    100       -
11   chain      14        11      140.800    100       -
12   chain      15        12      140.800    100       -
---------------------- SYROPE IC --------------------------------------
Line(s)   Tmax0    Tmean0
(-)        (N)       (N)
7,8,9    3.53e6    1.18e6
---------------------- OPTIONS -----------------------------------------
2        writeLog  - Write a log file
9.81     g         - gravitational acceleration
1025.0   WtrDnsty  - water density (kg/m^3)
0.2e-3   dtM       - time step to use in mooring integration (s)
3.0e6    kbot      - bottom stiffness (Pa/m)
3.0e5    cbot      - bottom damping (Pa-s/m)
243.3    WtrDpth   - water depth
0.1      dtIC      - time interval for analyzing convergence during IC gen (s)
100.0    TmaxIC    - max time for ic gen (s)
4.0      CdScaleIC - factor by which to scale drag coefficients during dynamic relaxation (-)
0.001    threshIC  - threshold for IC convergence (-)                    
---------------------- OUTPUTS -----------------------------------------
LINE7TENB
LINE8TENB
LINE9TENB
------------------------- need this line -------------------------------------- 
```
where the syrope.dat reads
```
../owc.dat  OWC     Original working curve lookup table path
LINEAR      WCType  Working curve formula: LINEAR, QUADRATIC, or EXP
3.39e8      k1      First parameter defining the working curve shape
0.0         k2      Second parameter defining the working curve shape
```

RyanDavies19:
That approach makes sense to me, and follows the approach of optional input file sections commonly seen in the MoorDyn-F / OpenFAST side of things (see https://moordyn.readthedocs.io/en/latest/inputs.html#external-loads-moordyn-f-only, https://moordyn.readthedocs.io/en/latest/inputs.html#failure-moordyn-f-only, https://moordyn.readthedocs.io/en/latest/inputs.html#control-moordyn-f-only).

Curious on others thoughts!

sanguinariojoe:
MoordynC has optional sections too, like rods and bodies.

I would keep the input file as uncluttered as possible. But I suppose there
is no much of a better approach here.

So for me you can go ahead

sanguinariojoe:
Well, it seems @zhilongwei has implemented the @mattEhall approach, so I suppose it is all green lights now. Tomorrow I'll check again for mem errors and if nobody has complained by then I will merge

sanguinariojoe:
@zhilongwei , before I merge, please add your test the memcheck-excluded ones:

https://github.com/FloatingArrayDesign/MoorDyn/blob/master/.github/workflows/memcheck.yml#L39

zhilongwei:
> @zhilongwei , before I merge, please add your test the memcheck-excluded ones:
> 
> https://github.com/FloatingArrayDesign/MoorDyn/blob/master/.github/workflows/memcheck.yml#L39

Have done so. Should I also rebase with dev again?

Issue #378, opened 2026-04-07
How to model a Failure condition with .txt


frozenheartsaber:
Hello

Sorry to bother again. I hope to know if I can simulate a failure condition with the MATLAB version? I see  (MoorDyn-F only) as shown in the documentation. And by the way, if I want to choose the attachment point as the failure position, aside from using generic numbered names like "Point 1" or "Point 2", is there a dedicated term for referring to this point?

Thank you

RyanDavies19:
Hi @frozenheartsaber, 

Unfortunately we haven't brought over the MD-F failures to MD-C yet, mostly because of different approaches to how the object states are stored internally. There is no dedicated term for referring to the failure points, your choice what to call it. The only requirement is that the indexing in the input files is sequential, i.e. the first failure you list has to have the ID failure 1 etc.

sanguinariojoe:
@RyanDavies19 , if you give me a so far accurate description I can provide you with all the infrastructure to do so.

It is my understanding that you need to create 2 free points on the breaking node position, and 2 lines representing the orig line subparts. Then you want to add the 2 lines and points and remove the orig line from the time integrator (i.e. the states).

RyanDavies19:
@sanguinariojoe, you pretty much have the general idea of it. Except that in MD-F failures are only possible at existing points, so rather than new lines being created (and deleted) just a single new point is added to the system. All the extra points are allocated during initialization, so in effect they are just 'activated' when a failure occurs. Failures can be tension or time triggered. 

This keeps the code quite simple, but does mean you need to know the failure location beforehand. This is pretty doable in most cases, because system weak points are well known. Here is the OpenFAST implementation: https://github.com/OpenFAST/openfast/pull/2214

There's a regression test in OpenFAST that can be used for comparison too.

sanguinariojoe:
OK, next week I build all the infrastructure to do that!

PR #379, opened 2026-04-16
Pointbreak

sanguinariojoe:
As promised, here you have the utilities to do lines breaking!

anim

It is just the scaffolding (+ an API entry). All the stuff about configurations, triggers and so on is not implemented. Thus it is not fully fixing https://github.com/FloatingArrayDesign/MoorDyn/issues/378, but some work has to be done yet


sanguinariojoe:
Hint: Check the [new test](https://github.com/core-marine-dev/MoorDyn/blob/pointbreak/tests/line_break.cpp)

sanguinariojoe:
Apparently the test fails on MacOs... Checking...

sanguinariojoe:
Everything is alright now. But I am finding some uninitialized values that might require a deeper analysis

sanguinariojoe:
Well, it is more elusive than I thought. Tomorrow I am finding and fixing it

sanguinariojoe:
@RyanDavies19 , no more errors! Please take a look, and if it suffices just merge and feel free to implement MoorDynF alike line breaking

RyanDavies19:
Hi @sanguinariojoe,

I haven't forgotten about this! Hoping to be able to look later this week/weekend. Thanks for putting it together.

RyanDavies19:
Thanks for adding it to the API, thats a great addition. Reading through the code, it seems like the remaining tasks would be:
1) IO through the MD input file still needs to be added in a way that matches OpenFAST
2) Accommodations for when lines are attached directly to rods (I don't recall if rod end nodes are in the point list or not)

Does that sound right?

Also a side note: I noticed in the animation you gave a point goes flying off w/ no lines. Is that intentional or does that show the point type changing from coupled/fixed to free? Based on the code I don't see where that would be happening but thought it would be good to double check.

RyanDavies19 (approved):

sanguinariojoe:
> I don't recall if rod end nodes are in the point list or not

Nope, they are not. Another `MoorDyn::BreakLine(moordyn::Rod* rod, moordyn::Line* line)` function is required.

Also, the C API entry shall be renamed to `MoorDyn_BreakLineAtPoint()` and another function `MoorDyn_BreakLineAtRod()` shall be created. Alternatively, we can create sort of an alias for generic entities, something like `MoorDynInstance`, as an alias of `void*`. But that would be way less explicit, letting the user to accidentally do strange things.

> Also a side note: I noticed in the animation you gave a point goes flying off w/ no lines. Is that intentional or does that show the point type changing from coupled/fixed to free? Based on the code I don't see where that would be happening but thought it would be good to double check.

As you can see here:

https://github.com/core-marine-dev/MoorDyn/blob/pointbreak/source/MoorDyn2.cpp#L809

The original point is kept, while the line is detached from it. That is indeed convenient in the case of coupled points (otherwise the user shall check and modify the number of DOFs he passes through `MoorDyn_Step()`). It would be risky for free points, but I actually do not see any border case on which a line can break and left a free point unconnected from all other entities. Maybe a big hanging weight?

Should we look for and remove standalone free points at the end of the function?

RyanDavies19:
@sanguinariojoe, 

I like the idea of two functions, one for rod detachments and one for point detachments. I also think we leave the possibility of free points with no lines, better to not unnecessarily constrain what can be simulated imo. 

My question was about this point moving away from the fairlead location (red arrow in screenshot below). It seems like the type is being set to free somehow, though I don't see where in the code that could happen. 
Screenshot 2026-05-11 at 8 14 07 AM

sanguinariojoe:
Nope, that one is the original coupled point

sanguinariojoe:
I mean, it is never set to free. It is still moving after the line break
because I am imposing a constant motion on x direction for the 3 coupled
points

RyanDavies19:
Makes sense!

RyanDavies19:
I am good with merging this if you are!

sanguinariojoe:
Done! Please, do not create a new release until we have the rod associated functions, so we do not need to add breaking changes

PR #386, opened 2026-06-08
fix(core): Return an error code when points below the seabed are detected

sanguinariojoe:
When a point falls below the seabed, Point.cpp:152 raises an exception, which is not catched, letting the full code to crash


RyanDavies19:
Thanks for catching this!

PR #398, opened 2026-07-17
feat: OpenFAST synchronization, docs updates, and VIV test fix

RyanDavies19:
**OpenFAST Synchronization**
Matching the code in OpenFAST, I added a new a line output channel (node accelerations, zero at end nodes) and a new rod out channel (node water velocities). I also added the waveKin_rampT ramp option (currents disabled during IC gen when ramping is active), which linearly ramps Water kinematics to prevent transients when starting up simulations. Finally, `!` and `%` are now accepted as comment characters alongside `#`.

**Documentation**
Documented the above changes, clarified per-node tension vs. FairTen/AnchTen, noted WEC-Sim/MATLAB can frequently cause crashes because of the WECSim implementation, and added VIV theory/validation paper reference. The last one documents the theory behind PR #295.

**Tests**
The existing VIV test is failing due to the OpenFAST synchronization changes causing a numerical phase shift in the output tensions (from lock-in occurring at different time during start-up). This is being picked up because the test's peak finding approach is too basic, and gets triggered by noise rather than the true signal. Thus, I improved VIV peak detection (longer run, later first check, min-peak-amplitude threshold for stability) to better pick up peaks. In doing that I made a new plotting script for visual validation, and thought it would be good to include if any future work needs debugging. The two figures below compare the results to the OpenFAST regression tests, and show that the new test peaks are still within the acceptable range. 
peaks
FFT


RyanDavies19:
@sanguinariojoe, requesting your review just so you are aware. There is nothing too major in here beyond an improved VIV test, and it is the code I used for the VIV paper (w/ validation study)

RyanDavies19:
Doxygen is missing the Body type in the detailed description
https://moordyn.readthedocs.io/en/master/doxygen/html/classmoordyn_1_1_body.html

https://github.com/FloatingArrayDesign/MoorDyn/pull/398/commits/2fc45a23b2bbc8b07e85ec3f7ed39c213de3c163 fixes this and adds the same type of documentation to all other objects.

sanguinariojoe:
I am on my vacation days. Next week I can check this out!

RyanDavies19:
@sanguinariojoe Enjoy the time off!

tests/Mooring/viv/plot_viv_timeseries.py
```diff
@@ -0,0 +1,340 @@
...
+    segment_time = time[mask]
+    segment_values = values[mask]
+    if segment_time.size < 4:
+        raise ValueError("Not enough samples after the FFT start time")
+
+    dt = float(np.median(np.diff(segment_time)))
+    uniform_time = np.arange(segment_time[0], segment_time[-1] + 0.5 * dt, dt)
+    uniform_values = np.interp(uniform_time, segment_time, segment_values)
+    centered_values = uniform_values - uniform_values.mean()
+    window = np.hanning(centered_values.size)
```
sanguinariojoe: You realize you are filtering abroad, right?

Image

I would do something like the following instead:

```
tail_length = centered_values.size // 10  # Or something else
window_mask = np.zeros(centered_values.size, dtype=bool)
window_mask[:tail_length] = True
window_mask[-tail_length:] = True
windowed_values = np.copy(centered_values)
windowed_values[window_mask] *= np.hanning(2 * tail_length)
```

i.e. limiting the filtering to the head and tail of the timeseries (FFT just asks for a periodic signal).

sanguinariojoe:
Hey!

I think this is good for merging.

I just thrown my two cents to you about the spectrum computation, which in my opinion is too aggressively smoothing the time series to make them periodic. Not very important though

sanguinariojoe:
The VIV test is failing on Windows, and just on Windows...

https://github.com/FloatingArrayDesign/MoorDyn/actions/runs/29873794709/job/88779751823?pr=398#step:14:107

This kind of things sometimes are related with the comma separator when traversing input files. Would you give it a look @RyanDavies19 ?

RyanDavies19:
The failing windows test is a little confusing, because it is reporting that a different period is being calculated. My guess is there is some accumulated floating point errors. I adjusted the test sensitivity slightly. But good to note in the future that there may be slight differences based on OS. 

Thanks for the pointers on the python script too, I revised it accordingly.

